PR #2, opened 2023-05-31
Remove unused folder and add eslint rules

amattu2:
### Overview

This PR removes an unused `/back` folder and introduces basic eslint rules.

PR #3, opened 2023-06-07
CRDCDH-80 Implement Unsaved Changes Prompt

amattu2:
### Overview

This PR covers the implementation of an "Unsaved Changes" prompt that invokes user action when leaving the current form section with unsaved changed.

Notably, when leaving a section with unsaved changes the following options become available to the user:

- Cancel – Cancel current action, hides prompt and stops invoked navigation action
- Save – Save the current form section then navigate onward
- Discard – Discards current changes and continues navigation

> **Note**: Any changes restored (i.e. Changing the Firstname field then restoring it back to the original value) will not count toward unsaved changes.

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
-import { useRoutes } from 'react-router-dom';
+import { RouterProvider, createBrowserRouter } from 'react-router-dom';
 import { ThemeProvider, CssBaseline, createTheme } from '@mui/material';
-import router from './router';
+import routeConfig from './router';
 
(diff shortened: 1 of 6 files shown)


amattu2:
> **Warning**: This PR is based on CRDCDH-33 and should not be merged before #2

amattu2:
TODO: Handle document unloads (i.e. non-react-originated redirects)

amattu2:
Pending font change.

jonkiky (approved):
LGTM!

PR #8, opened 2023-06-20
CRDCDH-95 Add apollo client and integrated with mock api server

jonkiky:
1. Added Apollo Graphql client 
2. Sample code for fetching data from mock api server
3. Code styling update -- auto correction


amattu2:
When writing the application type I accidentally wrote `id` instead of `_id`. Is it possible to fix the type in this PR? Everything else works as expected.

https://github.com/CBIIT/crdc-datahub-ui/blob/a5416df9b04ac1d6148d4244772e9c7f65b31326/src/types/Application.d.ts#L2

https://github.com/CBIIT/crdc-datahub-ui/blob/a5416df9b04ac1d6148d4244772e9c7f65b31326/src/content/questionnaire/FormView.tsx#L39-L40

Alejandro-Vega (approved):
LGTM. Works well on my end.

PR #9, opened 2023-06-22
CRDCDH-35 Redesign Form Section B

Alejandro-Vega:
**Overview**
This PR is a redesign of Form Section B for the ticket [CRDCDH-35](https://tracker.nci.nih.gov/browse/CRDCDH-35) The original implementation for the section was created by @amattu2. 

- Modified spacing and positioning of Form B
- Updated styling and spacing for every type of input (textfield, select, autocomplete, etc.)
- Updated styling and spacing for each section of the form
- Added section for Planned Publications
- Imported icons (dropdown arrows, and calendar) and added them as adornment to corresponding inputs
- Created DatePickerInput and **added Dayjs dependency** necessary for the DatePicker MUI component. If we want to switch date library adapters it should be addressed.


jonkiky (approved):
LGTM! --  We may need to think about how to improve the way to styling.

Alejandro-Vega:
@jonkiky I agree, we should find a better solution for styling and move common/repeated styling into global theme. The inputs for example all have very similar styling.

PR #10, opened 2023-06-23
Fix incorrect Application type

amattu2:
### Overview

This resolves an issue with the Application type in which the key `id` is used instead of the GQL defined `_id`.


Alejandro-Vega (approved):
LGTM

PR #13, opened 2023-06-26
Crdcdh 81 82

mtangmt:
**Overview**
This PR adds the NCI branded header and footer. This is copy pasted from the CCDI header with slight styling changes to fit what designers gave us, and code changes to satisfy typeScript

- class -> className where applicable for typeScript
- widths are 1400px instead of 1420
- Government header banner with the USA flag at the top
- This is pending changes for what will fill out the actual content of the header and footer from the feds.


Alejandro-Vega:
@mtangmt  We should definitely remove all of the extra assets that we will not use in this project. It is also reimporting fonts so maybe we need to rethink on how we would like to handle that. We also aren't using bento in this project so should we move it into the content folder instead?

amattu2:
Nice work @mtangmt! Just a few relatively minor changes:

1. Can you merge the latest from `origin/dev` into your branch and resolve any conflicts? 
2. After merging `origin/dev` into your branch, there will be (newly added) linting errors/warnings that will need to be addressed
3. Can we delete the [src/back](https://github.com/CBIIT/crdc-datahub-ui/tree/CRDCDH-81-82/src/back) folder? It was for something that we don't need anymore
4. I think @Alejandro-Vega captured this already, but let's go ahead and move the [src/bento](https://github.com/CBIIT/crdc-datahub-ui/tree/CRDCDH-81-82/src/bento) configs into [src/config](https://github.com/CBIIT/crdc-datahub-ui/tree/dev/src/config) instead. The `src/bento` folder is just a custodian configuration folder concept from the [Bento project](https://github.com/CBIIT/bento-frontend). 
5. I see `<CssBaseline>` was added to [src/layouts/index.tsx](https://github.com/CBIIT/crdc-datahub-ui/blob/CRDCDH-81-82/src/layouts/index.tsx), we do have it in a higher level element ([src/App.tsx](https://github.com/CBIIT/crdc-datahub-ui/blob/dev/src/App.tsx)) already so that can be removed.

jonkiky (approved):
.vscode folder is not necessary.  should included in git ignore settings.

PR #24, opened 2023-07-03
Adding underline styles, and fixing dropdown style

mtangmt:
This pr addresses requirements from Yizhen about the navbar underline tracking current navigation.

- The navbar section where you are navigated to a subsection in, will be underlined blue now.

Also addresses an issue brought up by Alex, where clicking a navbar section would shift the navbar.

- This no longer happens, and each section is static.

### DIFF
--- a/src/config/globalHeaderData.tsx
+++ b/src/config/globalHeaderData.tsx
@@ -52,24 +52,24 @@ export const navbarSublists = {
     },
     {
       name: 'CRDC Intake Applications',
-      link: '/#crdcia',
+      link: '/crdcia',
       className: 'navMobileSubItem',
     },
 ],
   About: [
     {
       name: 'Other Resources',
-      link: 'oa',
+      link: '/or',
       className: 'navMobileSubTitle',
     },
     {
       name: 'Cancer Genomics Cloud',
-      link: '/#cgc',
+      link: '/cgc',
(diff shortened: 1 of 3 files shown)


src/config/globalHeaderData.tsx
```diff
@@ -52,14 +52,14 @@ export const navbarSublists = {
...
-      link: '/#cgc',
+      link: '/#crdcia',
       className: 'navMobileSubItem',
     },
 ],
   About: [
     {
       name: 'Other Resources',
-      link: '',
+      link: 'oa',
```
amattu2, line 62: Is this intentionally a relative link? On the [questionnaire page](http://localhost:3000/questionnaire/1234) it links to `.../questionnaire/oa`

amattu2 (changes requested):
When clicking a sub-menu link, it reloads the entire application. My assumption is somewhere an anchor element is being used directly instead of the React Router `<Link>` component

I assume we should also remove the hash (`#`) prefixes from the header links since we aren't using a hashrouter. 

Everything else LGTM.

src/config/globalHeaderData.tsx
mtangmt: Uhh, no. Was a typo. Should be /or my bad.

jonkiky (approved):
LGTM!

PR #29, opened 2023-07-05
Adding ids to interactable items in header and form

mtangmt:
For QA reasons, we need IDs on interactable components (inputs and buttons). I added ID's to these components in the header and the questionnaire content.


jonkiky (approved):
LGTM!

PR #31, opened 2023-07-06
Form detecting false positive changes

Alejandro-Vega:
**Overview**
This PR is to fix the "Unsaved Changes" dialog from appearing when switching between form sections even when there hasn't been any new changes made.

### DIFF
--- a/src/components/StatusBar/utils.ts
+++ b/src/components/StatusBar/utils.ts
@@ -1,12 +1,21 @@
+import dayjs from "dayjs";
+
 /**
- * Format a date string to M/D/YYYY
+ * Format a date string to specified pattern
  *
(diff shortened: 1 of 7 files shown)


src/components/StatusBar/utils.ts
amattu2, line 1: If we're using the FormatDate function outside of the StatusBar component, we should probably move the location of the function. Maybe we should just merge all of the form-related utils files into one generic place?

Alejandro-Vega: @amattu2 Good catch, moved the utility to a more shared location.

amattu2 (approved):

jonkiky (approved):

PR #35, opened 2023-07-07
Crdcdh 128

mtangmt:
This implements the search functionality for static page content.

- When making a static page, please add all text in the StaticPageContentConfig.ts file, as this is what the search uses to search.
- We still have to decide on router links. Now, when we update links we have to update router.tsx, globalHeader.tsx, and  StaticPageContentConfig.ts


jonkiky (approved):
LGTM

PR #36, opened 2023-07-10
Crdcdh 135

mtangmt:
This implements ticket 135 with some adjustments.

- Added graphql values to switch input, so it no longer adds label name to the array to represent being switched on.
- Added new epidemiologic or cohort data type
- Added new file type extension column in the file type table depending on file type value.
- Added other file/extension grey text styling and custom validation in the file type table.


Alejandro-Vega:
@mtangmt Please update to merge to dev branch instead of main

mtangmt:
good catch!

jonkiky (approved):

PR #38, opened 2023-07-10
CRDCDH-80 Hide validation when saving from dialog

Alejandro-Vega:
**Overview**
This PR is to hide validation when saving from unsaved changes dialog. Also, I fixed issue with Cell Line & Model Systems to persist state as the value was changed to initial value when dialog appeared.


mtangmt:
known Cell Line & Model Systems checkbox issue to be fixed in #37

PR #42, opened 2023-07-12
CRDCDH-133 Updating Program/Study 

Alejandro-Vega:
**Overview**
This PR is to add name and abbreviation to the label for program and study Autocomplete inputs. 

- Updated Program/Study logic
- Added name and abbreviation to input labels
- Modified autocomplete to accept generic types as values and options
- Update autocomplete styling to use styled instead
- Fixed initialValues and applicationData not doing a deep merge causing some nested properties to be misssing
- Fixed toggleContent being conditionally rendered causing for the input to not appear in the DOM and form parser removing the property from values causing a detected change when navigating through sections
- Fixed textInput not allowing for empty string values
- Added key to autocomplete inputs to fix issue cause by study depending on program
- Updated study names and abbreviations
- Added Study Description tooltip


amattu2 (commented):
When saving on Section B (without making any changes), I get an error for the dbGaPPHSNumber field. ~~If I interact with the switch before saving, it doesn't occur.~~ Correction...it occurs only when the switch is off.

image

It doesn't seem to happen on the dev/mvp branch, but if this is unrelated to your changes, we can disregard this issue.

Alejandro-Vega:
@amattu2  Good catch! It was caused when a hidden input also has a required property, but since the field should be required, then I had to conditionally add that property. I didn't add this field, but was fixing a different problem caused by it which made the unsaved changes dialog come back. Thanks.

src/components/Contexts/FormContext.tsx
```diff
@@ -124,15 +125,16 @@ export const FormProvider: FC<ProviderProps> = (props) => {
     // Update the state when the lazy query response changes
     if (data) {
       const applicationData = data?.getApplication;
+      const newData: Application = {
+        ...merge(initialValues, applicationData),
```
amattu2, line 129: Can we swap the argument order for this Lodash merge? It's updating the InitialValues variable which will combine values from different forms as you go through them.

image

Alejandro-Vega: Lodash is weird. I updated to make a deep copy of the initialValues object like you suggested. Thanks Alec!

amattu2 (approved):
Everything LGTM. 

Nice work including new test cases also!

PR #43, opened 2023-07-12
CRDCDH-39

JoonLeeNIH:
Uses the authorization code from NIH to log in to AuthN. Jira ticket [CRDCDH-39](https://tracker.nci.nih.gov/browse/CRDCDH-39)


src/components/Contexts/AuthContext.tsx
```diff
@@ -0,0 +1,84 @@
...
+  useState,
+} from "react";
+import {
+  isLoggedIn,
+  logIn
+} from '../../api/authn';
+
+export type ContextState = {
+  isLoggedIn: boolean;
+  user: object;
```
amattu2, line 16: Is it possible we could create a type for the user object? This will make it easier to predict the data structure provided by the AuthContext for components using it.

e.g.
```
export type User = {
  firstName: string;
  role: "User" | "FederalLead"; // any other roles supported
  // ... other fields provided by the API
};
```

JoonLeeNIH: Sounds good, Alec. I know I'll miss more stuff like this as I get used to TypeScript, so please feel welcome to point things out as you see them.

jonkiky (approved):

PR #49, opened 2023-07-18
CRDCDH-161 and 103

mtangmt:
This implements CRDCDH-161 which changes the header.
Changes include: 
- Combining mobile and tablet header as they are the same code.
- Ensuring blue underline on header work on link header items as well as clickable sublist items.
Also implements ticket 103 which requires login for navigating to CRDC Submission Requests from the navbar.


mtangmt:
Wait for CRDCDH-149 to be merged, as this works with that.

src/content/login/Controller.tsx
```diff
@@ -9,12 +10,14 @@ const loginController = () => {
...
-
+  const { state } = useLocation();
+  const redirectURLOnLoginSuccess = state && state.redirectURLOnLoginSuccess ? state.redirectURLOnLoginSuccess : null;
   const urlParam = {
     client_id: `${NIH_CLIENT_ID}`,
     redirect_uri: `${NIH_REDIRECT_URL}`,
     response_type: 'code',
-    scope: 'openid email profile'
+    scope: 'openid email profile',
+    state: redirectURLOnLoginSuccess
```
amattu2, line 20: Can we add the state param only if it's not null? When logging in without using the dialog, we're getting redirected to `[...url...]/null` which results in 404.

jonkiky (approved):

PR #50, opened 2023-07-18
Crdcdh 149

mtangmt:
This PR Implements ticket 149, and ticket 67 which shows the user's name when logged in.

- When user is logged in, name shows up, with user profile and logout options.
- Mobile design is implemented.

**Waiting for authContext so these components can correctly distinguish when user is logged in**


src/components/Header/components/NavbarDesktop.tsx
```diff
@@ -341,15 +406,43 @@ const NavBar = () => {
             })
           }
         </UlContainer>
-        <StyledLoginLink id="header-navbar-login-button" to="/login">
-          Login
-        </StyledLoginLink>
+        {testLoggedInBool
+          ? (
+            <LiSection>
+              <div style={{ margin: "0", minWidth: "150px", border: "none" }} className={clickedTitle === testData.name ? 'navTitleClicked' : 'navTitle'}>
```
jonkiky, line 412: to avoid inline style as much as possible.  It is unsafe. 
https://www.iothreat.com/blog/csp-style-src-unsafe-inline.

jonkiky (approved):

PR #71, opened 2023-08-03
CRDCDH-54 Misc changes to Listview

amattu2:
### Overview

- **This PR primarily updates the ListView to create an empty application before redirect**. 
- Adds the new column `Last Updated Date` based off of the `updatedAt` field to the Submission Request List table per new requirements.
- This also will fix an issue with MVP-2 showing incorrect buttons to Org Owners.


amattu2:
NOTE: Need to disable query caching. When creating an application and returning to the listing, the old query is cached and the new app is not visible until reload.

jonkiky (approved):

PR #79, opened 2023-08-04
Crdcdh 45 validation bugfix

mtangmt:
This fixes a bug in section D, where the form would not be marked complete because of a incorrectly setting validation on an invisible component.
Now the section D form should save properly, even when the imagingDeIdentified component is invisbile.

### DIFF
--- a/src/components/Questionnaire/RadioYesNoInput.tsx
+++ b/src/components/Questionnaire/RadioYesNoInput.tsx
@@ -120,7 +120,7 @@ const RadioYesNoInput: FC<Props> = ({
     setVal(newValue === "" ? null : newValue);
   };
   useEffect(() => {
-    if (required && !val) {
+    if (required && val === null) {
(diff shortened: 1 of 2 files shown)


amattu2 (changes requested):
Can you confirm that we need to add this new prop? A similar issue came up on Section B and the solution was conditionally setting the `required` prop on the nested element.

https://github.com/CBIIT/crdc-datahub-ui/blob/1b7f04ee3112f05295cd241410032a00260b2b7a/src/content/questionnaire/sections/B.tsx#L399

https://github.com/CBIIT/crdc-datahub-ui/blob/1b7f04ee3112f05295cd241410032a00260b2b7a/src/content/questionnaire/sections/B.tsx#L411

mtangmt:
That's smarter.

jonkiky (approved):

PR #90, opened 2023-08-10
CRDCDH-49 Form Auth Update

Alejandro-Vega:
**OVERVIEW**
This PR aims to expand the amount of variables to calculate the form mode, rather than just "User" and "FederalLead" roles and form status. I separated the logic for calculating the form mode into its own utility file where I could create tests for it easier and also simplify the useFormMode hook itself.

- Added organization to getMyUser query and User type
- Separated useFormMode hook logic into separate utility file to make it testable
- Created detailed tests for each role, orgRole, formStatus, who submitted the form, which organization the submission belongs to, which organization the user belongs to (if any), etc.
- Updated Approve/Reject form dialog to maintain internal state for review comment and only pass the value onSubmit. Previously, it was causing the parent to re-render due to state changes. 

**TICKETS**
[CRDCDH-49](https://tracker.nci.nih.gov/browse/CRDCDH-49)

**NOTE**
This implementation assumes a user will only belong to a single organization.

### DIFF
--- a/src/types/Application.d.ts
+++ b/src/types/Application.d.ts
@@ -158,12 +158,3 @@ type Organization = {
   _id: string;
   name: string;
 };
-
-type OrgInfo = {
-  orgID: string;
-  orgName: string;
-  orgRole: "Owner" | "Submitter" | "Concierge"; // Concierge can only be assign to a Curator
-  orgStatus: "Active" | "Inactive" | "Disabled";
-  createdAt: string; // 2023-05-01T09:23:30Z, ISO data time format
-  updateAt: string; // 2023-05-01T09:23:30Z  ISO data time format
-};
--- a/src/types/Application.d.ts
+++ b/src/types/Application.d.ts
@@ -158,3 +158,12 @@ type Organization = {
   _id: string;
   name: string;
 };
+
+type OrgInfo = {
+  orgID: string;
+  orgName: string;
+  orgRole: "Owner" | "Submitter" | "Concierge"; // Concierge can only be assign to a Curator
+  orgStatus: "Active" | "Inactive" | "Disabled";
+  createdAt: string; // 2023-05-01T09:23:30Z, ISO data time format
+  updateAt: string; // 2023-05-01T09:23:30Z  ISO data time format
+};
--- a/src/utils/formModeUtils.ts
+++ b/src/utils/formModeUtils.ts
@@ -1,0 +1,158 @@
+export type FormMode = "Unauthorized" | "Edit" | "View Only" | "Review";
+export type FormStatus = "New" | "Submitted" | "In Review" | "Approved" | "In Progress" | "Rejected";
+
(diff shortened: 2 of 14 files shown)


src/utils/formModeUtils.ts
```diff
@@ -0,0 +1,158 @@
...
+const _getFormModeForFederalLead = (
+  user: User,
+  data: Application
+): FormMode => {
+  if (user?.role !== "FederalLead") {
+    return FormModes.UNAUTHORIZED;
+  }
+
+  const { status: formStatus } = data || {};
+  const isStatusViewOnlyForLead = ["In Progress", "Approved", "Rejected"].includes(formStatus);
```
amattu2, line 66: Federal Leads should basically never be unauthorized to view at a minimum. For now, the `New` status should be added here.

src/utils/formModeUtils.ts
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * @param {User} user - The current user
+ * @param {Application} data - The current application/submission
+ * @returns {FormMode} - Updated form mode based on role, organization role, and form status
+ */
+export const getFormMode = (
+  user: User,
+  data: Application,
+): FormMode => {
+  if (!user || (!user?.role && !user?.organization?.orgRole) || !data) {
```
amattu2, line 126: A user should always have a role. This scenario should never occur, but we may want to consider missing `user.role` as unauthorized, even if they have an organizational role.

amattu2 (changes requested):
@Alejandro-Vega Very solid work here with the refactoring and comprehensive test cases. I added some comments which are optional, and one change required for the Fed lead role.

src/utils/formModeUtils.ts
```diff
@@ -0,0 +1,158 @@
+export type FormMode = "Unauthorized" | "Edit" | "View Only" | "Review";
+export type FormStatus = "New" | "Submitted" | "In Review" | "Approved" | "In Progress" | "Rejected";
```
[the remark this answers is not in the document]
Alejandro-Vega: Great catch! Wasn't even using it either. Removed it.

src/types/Application.d.ts
```diff
@@ -158,3 +158,12 @@ type Organization = {
   _id: string;
   name: string;
 };
+
+type OrgInfo = {
```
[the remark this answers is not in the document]
Alejandro-Vega: Moved it, thanks!

src/utils/formModeUtils.ts
Alejandro-Vega: Removed the check entirely and set default fallback to view only.

src/utils/formModeUtils.ts
Alejandro-Vega: Updated to return unauthorized when user role does not exist. Thanks.

amattu2 (approved):

PR #92, opened 2023-08-11
fixing positive integer input

mtangmt:
This pr fixes a bug where non-numerical characters could be typed into the file count input in the table in section D.
Now, only positive integers can be typed into the input, as requried.


jonkiky (approved):
LGTM!

PR #98, opened 2023-08-22
Crdcdh 65

mtangmt:
This implements ticket 65.
To test, I modified either:
- thresholdTime on line 124 of InactivityDialog.tsx, to be 1800
- modified line 153 to be `const ttl  = data.ttl - 1790;`

Modifying the threshold time to be 1800, means the warning dialog will pop up at 30 minutes left in the time (immediately).
Modifying the ttl received from the backend leaves only 10 seconds left (1800 - 1790 = 10) on the timer and then we can test both the warning dialog, and the session timeout.
To test the extension of the ttl, just console.log the ttl recieved from the backend.


mtangmt:
After discussing with Kailing and updating the ticket requirements. We decided that the dialog pop-up made more sense for timeout, and that having a banner and the dialog was redundant. 
So now, on timeout, the user is logged out, redirected to home page/login page, and only the dialogbox shows up.

jonkiky (approved):

PR #114, opened 2023-09-07
CRDCDH-336 Monitor for freeSolo input

amattu2:
### Overview

This PR fixes bug [CRDCDH-336](https://tracker.nci.nih.gov/browse/CRDCDH-336) which causes any `AutocompleteInput` to not validate custom user input.

To replicate the bug:

1. Find a autocomplete input with no value currently set (Important)
2. Type custom text into the field
3. Save
4. See the validation error

The fix:

- Updated the autocomplete input to watch for input within the text field. Any changes in the text field triggers the existing processValue function.


amattu2:
Waiting for the bug ticket to be assigned to me.

jonkiky (approved):

PR #122, opened 2023-09-14
require at least one data type

mtangmt:
This implements ticket 366, which requires at least one data type to be selected as yes, or that "Other data types" has characters in it, in section D.


Alejandro-Vega (approved):
LGTM

PR #138, opened 2023-09-22
Adding tooltips to show errors for file type table

mtangmt:
This implements ticket 397. 
It solves the number of files field by putting a cap on the length allowed to be input (10 chars).
I solved the error issue kailing asked by making a custom tooltip, but need approval for it.


mtangmt:
Kailing approved

PR #155, opened 2023-09-28
CRDCDH-449 Allow additional saves when duplicate Study Abbrev error

Alejandro-Vega:
**OVERVIEW**
This PR aims to allow for additional saves to be made even when no changes have been made and the duplicate Study Abbreviation error has occurred. 

- Displays alert banner when unable to save due to duplicate Study Abbreviation
- Prevent user from navigating to another section with Next/Back buttons as well as Progress bar buttons, when user has duplicate Study Abbreviation error 
- Fixed small bug where the "Next" button in section D wouldn't enable when all sections complete in a new submission

**TICKETS**
[CRDCDH-449](https://tracker.nci.nih.gov/browse/CRDCDH-449)

### DIFF
--- a/src/components/Contexts/FormContext.tsx
+++ b/src/components/Contexts/FormContext.tsx
@@ -154,7 +154,6 @@ export const FormProvider: FC<ProviderProps> = ({ children, id } : ProviderProps
         questionnaireData: data
       }
     };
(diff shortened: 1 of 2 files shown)


amattu2 (commented):
Two questions/comments added. Everything else looks good to me.

PR #181, opened 2023-10-25
Crdcdh 525

mtangmt:
This pr adds CCDI as an option for data commons parameter when creating a data submission, as required in ticket 525.


amattu2 (changes requested):
The additional CCDI option looks good, but it's not selecting CDS by default (a new requirement).

amattu2 (approved):
LGTM. Nice work.

PR #186, opened 2023-10-30
CRDCDH-445 API Token

Alejandro-Vega:
**OVERVIEW**
This PR aims to add an "API Token" option under the user navigation dropdown where they can generate and copy their token. 

- Only Submitter and Org Owner roles may view it.
- Previous token should disappear after dialog closes

**TICKETS**
[CRDCDH-445](https://tracker.nci.nih.gov/browse/CRDCDH-445)


amattu2 (approved):
Solid work. Approved

PR #191, opened 2023-11-01
Crdcdh 480 517

mtangmt:
This PR aims to implement tickets 480 and 517, which are rejecting, and inquiring on submission requests.
These do not fulfill the ticket requirements completely, as email notifications still need to be done on the backend, but those cannot be implemented by the frontend.
This should fulfill all frontend requirements.


amattu2 (changes requested):
Really solid work here @mtangmt! A few action items:

1. Looks like there's a leftover TODO comment
2. A test case for formModeUtils is failing. See below.
3. See my optional comment for adding more form mode test cases

This test case is failing. Can you look into it? `npm run test` will help identify the reason.

https://github.com/CBIIT/crdc-datahub-ui/blob/bf07f8d75d64b2f3bef371b0eca9bc061f53b180/src/utils/formModeUtils.test.ts#L129-L135

___

Optional, but I would love to see if we extend the current test suite for formModeUtils to cover the new status that was added? Examples of potentially useful cases to test for (if they don't exist already): 

- Rejected is a final state, no one should be able to do anything beyond that
- Inquired should be read only for Fed Lead but not for the submission request owner
- ...Anything else you think we should cover?

amattu2 (approved):
LGTM. Awesome work.

PR #192, opened 2023-11-01
CRDCDH-377 Submission download validation logs

Alejandro-Vega:
**OVERVIEW**
This PR aims to retrieve and download validation logs in the Data Submission Dashboard > Quality Control tab.
- It will only retrieve logs if you own the submission
- It will un-disable the download buttons if logs exist for either metadata or file
- Process is: 
  - Retrieve Logs with download URLs if submission owner
  - Download clicked
  - Retrieve logs for the specific type
  - Zip up the logs and download for user
- Removed old Mock API query

**TICKETS**
[CRDCDH-377](https://tracker.nci.nih.gov/browse/CRDCDH-377)

**NOTE**
- There is currently no way for validation logs to be created
- For testing in DEV2, use submission `9f42b5f1-5ea4-4923-a9bb-f496c63362ce` which will retrieve a log, you must also make yourself the submission owner
- This is only to download logs and not CSV files. ([CRDCDH-554](https://tracker.nci.nih.gov/browse/CRDCDH-554) will make the update)
- jszip and file-save dependencies added


src/content/dataSubmissions/QualityControl.tsx
```diff
@@ -0,0 +1,212 @@
...
+            alignItems="center"
+          >
+            <StyledDownloadLabel
+              variant="body1"
+              textAlign="end"
+              sx={{ color: !metadataLogs?.length ? "#AAA" : "083A50" }}
+            >
+              Metadata Validation Results
+            </StyledDownloadLabel>
+            <StyledDownloadButton
```
amattu2, line 185: For these two download buttons, one change I would suggest is to throw it in the loading state while downloading/zipping the files. If you have slow internet, it may look like nothing is happening. Clicking fast also downloads multiple times, which may not be desirable. 

This can wait until [CRDCDH-554](https://tracker.nci.nih.gov/browse/CRDCDH-554).

amattu2 (commented):
One suggestion added, but it can definitely wait until CRDCDH-554.

Can you also confirm whether the QC tab should be accessible even if you don't own the submission? The ticket doesn't seem to say one way or the other.

src/content/dataSubmissions/QualityControl.tsx
Alejandro-Vega: Added state to set button to loading while downloading. As well as a missing try/catch

amattu2 (approved):
LGTM!

Will revisit the QC tab access later if necessary.

PR #195, opened 2023-11-03
fixing router bug

mtangmt:
This pr addresses the bug outlined in 568.
The bug arose because the router was rendering the page regardless of authentication status (and then navigating to home), and the data-submissions page required information that authenticated users have.
This adds logic to make sure the page is only rendered if an authenticated user is clicking on the link.


amattu2 (approved):
LGTM. Thanks for addressing this!

PR #199, opened 2023-11-07
implement gov shutdown banner

mtangmt:
### Overview
This implements ticket 547 which is the government shutdown banner.
Adds the government shutdown banner script to our index.html
To test, just change the URL in the include-html src to "https://cbiit.github.io/crdc-alert-elements/banners/government-shutdown-test.html"

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -30,8 +30,14 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>CRDC DataHub</title>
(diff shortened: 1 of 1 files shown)


amattu2 (changes requested):
Implementation LGTM and works as expected. We do need to update the script and banner URLs to the ones from the ticket.

`nci-softwaresolutions-elements` is for a separate team.

amattu2 (approved):

PR #203, opened 2023-11-08
CRDCDH-575 moving icons

mtangmt:
### Overview
This PR implements 575, but differently since dev couldn't load the icons correctly.
### Implementation Details
This moved all the history icons into assets so hopefully dev can load the icons correctly.


amattu2 (approved):
Seems to work as expected. Nice job!

PR #207, opened 2023-11-09
Crdcdh 586

mtangmt:
### Overview
This PR addresses the bug in ticket 586 where the date picker would incorrectly show the error border on the current date.
This bug was caused by the date picker returning the incorrect midnight for the day due to daylight savings.
To fix this, I just applied a time zone to the value returned by the date picker, and now it should work.
To test this, I changed the time zone for both node, and the browser following these steps: https://www.browserstack.com/guide/change-time-zone-in-chrome-for-testing. 
If you console.log the value in the onChange function prop to to StyledDatePicker, you can see that the value depends on the browser time zone.
This leads me to believe the MUI date picker uses the browser's time zone, so by normalizing the time zone in the code, I believe this issue is fixed.
### Related tickets
https://tracker.nci.nih.gov/browse/CRDCDH-586


amattu2 (commented):
LGTM. Waiting on @Alejandro-Vega's review then we can merge.

Alejandro-Vega (approved):
LGTM!

PR #240, opened 2023-12-12
CRDCDH-667 Clear history when populating model

amattu2:
### Overview

This PR resets the Model Navigator search history any time the redux store is repopulated. A future solution would be to isolate the model's search history and not clear it automatically, but that will require DMN core updates.

- Dispatch the redux event `SEARCH_CLEAR_HISTORY` when populating the model redux store

### Related Ticket(s)

CRDCDH-667


Alejandro-Vega (approved):
LGTM

PR #249, opened 2023-12-20
CRDCDH-659 Poll for file validation status

amattu2:
### Overview

This PR introduces data submission polling to track the file validation status. 

> [!TIP]
> The validation API works, but no validation happens yet; thus you will need to manually update the data submission object to change `metadataValidationStatus` (sp.) or `fileValidationStatus` (sp) from "Validating" to another state in order to see the button re-enable.

- Removed the usage of states for the Data Submission and Stats so that we can use native Apollo polling directly
- Migrated from `useLazyQuery` to `useQuery` for `GET_SUBMISSION` query
  - This will auto fetch the submission on component mount so that we don't need to manually initiate it
- Fixed a 508 issue with an empty button when there are no batch errors
- Fixed GenericAlert hiding under the Navbar Dropdown container (had equal z-index)

Screenshot 2023-12-22 at 9 35 43 AM
 
### Related Ticket(s)

CRDCDH-659

> [!IMPORTANT]  
> This ticket can be merged, but needs to go back to On Hold while we wait for the validation service.

### DIFF
--- a/src/components/DataSubmissions/ValidationControls.tsx
+++ b/src/components/DataSubmissions/ValidationControls.tsx
@@ -1,4 +1,4 @@
-import { FC, useMemo, useState } from 'react';
+import { FC, useEffect, useMemo, useState } from 'react';
 import { useMutation } from '@apollo/client';
 import { FormControlLabel, RadioGroup, styled } from '@mui/material';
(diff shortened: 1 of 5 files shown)


Alejandro-Vega (approved):
LGTM!

PR #253, opened 2024-01-02
CRDCDH-594 Submit Using Validation Status

Alejandro-Vega:
### Overview

This PR aims to use `metadataValidationStatus` and `fileValidationStatus` properties from the submission, instead of the stat nodes, to decide whether or not to disable the submit button. This is so that the FE and BE are both on the same page.

N/A

### Related Ticket(s)

[CRDCDH-594](https://tracker.nci.nih.gov/browse/CRDCDH-594)
[CRDCDH-635](https://tracker.nci.nih.gov/browse/CRDCDH-635)


amattu2 (changes requested):
This logic seems to incorrectly allow for submission in this scenario:

|metadataValidationStatus|fileValidationStatus|
|:-:|:-:|
|`null`|`Passed`|

My understanding is that, at a minimum, all submissions need metadata? It may not be possible for this scenario to occur in a real submission, but I would recommend we cover for it.

___

Other scenarios tested and working as expected:

|metadataValidationStatus|fileValidationStatus|
|:-:|:-:|
|`Passed`|`null`|
|`Warning`|`null`|
|`Passed`|`Passed`|
|`Passed`|`Warning`|
|`Warning`|`Passed`|

Alejandro-Vega:
> This logic seems to incorrectly allow for submission in these scenario(s):
> 
> metadataValidationStatus	fileValidationStatus
> `null`	`Passed`
> `null`	`null`
> My understanding is that, at a minimum, all submissions need metadata? It may not be possible for this scenario to occur in a real submission, but I would recommend we cover for it.
> 
> Other scenarios tested and working as expected:
> 
> metadataValidationStatus	fileValidationStatus
> `Passed`	`null`
> `Warning`	`null`
> `Passed`	`Passed`
> `Passed`	`Warning`
> `Warning`	`Passed`

Updated to disable submit if metadataValidationStatus is null

PR #254, opened 2024-01-03
CRDCDH-661 QC Results Filters

amattu2:
### Overview

This PR adds Data Submission Quality Control results filters to the QC Results tab

- Add 3 filters to the QC tab
- Migrate the listBatches query schema to a set of fragments so the QC filters can reuse the same query

### Related Ticket(s)

CRDCDH-661


Alejandro-Vega (changes requested):
There seems to be an issue when you set the rows per page to something smaller like 5, then you select a filter that doesn't have any results like Severity "Warning", then switch the filter back to "All". The table will show more than just 5 rows.

amattu2:
> There seems to be an issue when you set the rows per page to something smaller like 5, then you select a filter that doesn't have any results like Severity "Warning", then switch the filter back to "All". The table will show more than just 5 rows.

Since I was using `handleFetchQCResults` directly, any filter change didn't have the contextual information on the current rows per page. Using `setPage` with `refresh` caused the query to run twice (f0f0d49), so I modified the `setPage` implementation to also force-refetch data (d935ffa). I'm open to any suggestions for an improved approach.

Alejandro-Vega (approved):
LGTM!

PR #255, opened 2024-01-03
CRDCDH-538 Data Submission QC Statistics (v2)

amattu2:
### Overview

This PR introduces changes to the CRDCDH-538 implementation. We're still waiting on the final decision for which design to go with, so this is based off of Design 1.

- Rename the chart sections (Primary/Secondary charts)
- Fix an issue with the charts reanimating occasionally when fetching the submission statistics. Caused by:
  - The submission stats API doesn't return the stats in a consistent order, so we're sorting them by the nodeName
  - Prevent rerendering of the ValidationStatistics and NodeChart by memoizing the entire component
- Redesign the legend and make it interactive
- Add the missing "Total XXX" count inside the primary chart when it's NOT being hovered over
- Make the node charts interactive when hovering over a specific slice
- Add interactive legend with filters

### Related Ticket(s)

CRDCDH-538


Alejandro-Vega (approved):
LGTM!

PR #264, opened 2024-01-16
CRDCDH-689 Warn on removal of study with active data submission

amattu2:
### Overview

This PR adds a confirmation dialog when an Admin attempts to remove a study currently tied to an active data submission.

- Implement confirmation dialog based on API Token / Uploader CLI dialog (per K.C. request)
- Modify getOrganization query to also fetch current data submissions tied to an org
  - NOTE: `getOrganization` is only used by this page. If we ever use it elsewhere, we will need to refactor it to not ALWAYS include the data submissions

### Related Ticket(s)

CRDCDH-689


Alejandro-Vega (approved):
LGTM!

PR #269, opened 2024-01-24
CRDCDH-754 Poll for Batch Status Updates

amattu2:
### Overview

This PR updates the `BatchStatus` type to conform to the current definition from CRDCDH-736 and implements listBatch polling if there are any batches in the Uploading state.

- Update TypeScript definition of BatchStatus per CRDCDH-736
- Modify listBatches GraphQL to query for full list of batch statuses
- Poll listBatches until there are no batches uploading, when there are no batches, update the data submission to retrieve latest stats

### Related Ticket(s)

CRDCDH-754


Alejandro-Vega (approved):
LGTM!

PR #280, opened 2024-02-12
CRDCDH-538 Validation Statistics Design #4

amattu2:
### Overview

This PR converts the implementation of Data Submission Statistics to Design 4.

- Change primary chart to a bar chart w/View by Count or Percentage (i.e. normalized data or not)
- Add custom tooltip for primary chart
- Add "Validation Status" text to legend and update LegendItem text color
- Move primary label to centered above the chart
- Migrate statistic type definitions to a independent type file
- Fix utility `index.ts` file not exporting the statisticUtils
- Potential fix for "Total 0" placeholder randomly disappearing
  - The fix assumes that it's impossible for a hoveredSlice to have a value of 0 because slices with a value zero don't appear in the chart
- Remove carousel infinite scroll (ba01a0e)

### Related Ticket(s)

CRDCDH-538


src/components/NodeTotalChart/CustomTick.tsx
```diff
@@ -0,0 +1,30 @@
+import { styled } from '@mui/material';
+
+const StyledLabel = styled('tspan')({
+  fontFamily: "Roboto",
```
Alejandro-Vega, line 4: Correct me if I'm wrong, but I don't think we have Roboto font imported from anywhere.  I don't see it via CDN, dep, or local. I added a `cursive` font fallback and it used the fallback font.

src/components/NodeTotalChart/index.tsx
```diff
@@ -33,57 +37,47 @@ const StyledChartContainer = styled(Box)({
...
-      </PieChart>
-    </StyledChartContainer>
-  );
-};
+const NodeTotalChart: FC<Props> = ({ data, normalize = true }) => (
+  <StyledChartContainer>
+    <ResponsiveContainer height="100%" width="100%">
+      <BarChart
+        layout="horizontal"
+        data={data}
```
Alejandro-Vega, line 45: The order of the data on the x-axis is different per design, ticket, and implementation. Not sure which is the correct way.

src/components/Carousel/index.tsx
```diff
@@ -69,11 +66,9 @@ const ContentCarousel: FC<Props> = ({ children, ...props }: Props) => (
       responsive={sizing}
       swipeable
       draggable
-      infinite
-      centerMode
       arrows
```
Alejandro-Vega, line 69: Carousel left/right arrows differ in design

Screenshot 2024-02-21 at 11 37 15 AM
Screenshot 2024-02-21 at 11 37 53 AM

src/components/NodeTotalChart/Tooltip.tsx
```diff
@@ -0,0 +1,88 @@
...
+  lineHeight: "18px",
+  background: "#EAF1F2",
+  padding: "0 4px",
+});
+
+type Props = {
+  normalized: boolean;
+} & TooltipProps<number, SeriesLabel>;
+
+const Tooltip: FC<Props> = ({ active, payload, label, normalized }: Props) => {
```
Alejandro-Vega, line 57: (Optional) Missing tooltip tail pointing at hovered slice in chart. In design it looks to dynamically change from left to right. Not sure how hard it would be to implement this due to Recharts.

Alejandro-Vega (changes requested):
Looks great! I just left some small comments on design changes and some optional ones.

src/components/NodeTotalChart/CustomTick.tsx
amattu2: Good catch. Fixed in 5094d72

src/components/Carousel/index.tsx
amattu2: Completely missed this, thanks for catching it. Updated to use the existing chevrons in 4fb4c3c.

src/components/NodeTotalChart/Tooltip.tsx
amattu2: It seems possible to implement based on the coordinates of where the tooltip is placed but it's a high-ish effort. Leaving this detail omitted ATM. Will revisit in the future.

Thanks for pointing this out!

Alejandro-Vega (approved):

PR #284, opened 2024-02-16
CRDCDH-835 Failed batch has no Node Types for files

amattu2:
### Overview

Minor bug fix to use a fallback placeholder for files without a node type. This occurs during `Uploading` or sometimes `Failed` Batch states.
 
- Use `N/A` when there is no nodeType associated with a file

### Related Ticket(s)

CRDCDH-835


Alejandro-Vega (approved):
LGTM!

PR #296, opened 2024-03-01
Data submission tabs alignment Hotfix

Alejandro-Vega:
### Overview

Fixed alignment of tabs not always being flush with the section under it causing the border from the section under it to show. The problem was only visible in certain scenarios such as zoomed in browsers or different screens.

- Updated alignment to flex-end so that it is always flush with the lower section

Screenshot 2024-03-01 at 4 44 09 PM
Screenshot 2024-03-01 at 4 44 28 PM

### Related Ticket(s)

N/A


amattu2 (approved):
Awesome catch. Looks much better now.

PR #297, opened 2024-03-01
fix: Data File not appearing in Node Type column

amattu2:
### Overview

This is a quick fix for CRDCDH-810, where the "Data File" value wouldn't appear in the Node Type column properly.

> [!NOTE]
> This submission can be used to observe the issue `8a6368a2-9066-4326-b4b2-3f845db83e68`

- Override nodeType property for validationType `data file`

### Related Ticket(s)

CRDCDH-810


amattu2:
BE will make the change. Minor renaming changes likely in the FE. Will be a separate PR.

PR #309, opened 2024-03-14
fix: Random AuthContext test case failing in CI

amattu2:
### Overview

This PR aims to fix the random failures of the `should update the localStorage cache when the user is verified` test case for the AuthContext. It appears to be caused by a race condition.

e.g. https://github.com/CBIIT/crdc-datahub-ui/actions/runs/8100132739/job/22137439227

- Wrap the localStorage expectation in a Jest waitFor action which will continue to "re-test" the expect for up to 1 second before failing

### Related Ticket(s)

N/A


Alejandro-Vega (approved):
LGTM!

PR #312, opened 2024-03-15
CRDCDH-951 Set default Data Submission tab when unspecified

Alejandro-Vega:
### Overview

Updated Data Submission Dashboard to default to `Data Activity` tab when the tab is missing in the URL. Also displaying workflow buttons on any tab.

- Set default tab to `Data Activity` when tab is undefined
- Updated workflow buttons to display them always, regardless of tab selected

### Related Ticket(s)

[CRDCDH-951](https://tracker.nci.nih.gov/browse/CRDCDH-951)
**NOTE:** Bug ticket to fix showing workflow buttons on any tab can't be created due to read-only Jira board, but was verbally discussed

### DIFF
--- a/src/content/dataSubmissions/DataSubmission.tsx
+++ b/src/content/dataSubmissions/DataSubmission.tsx
@@ -6,7 +6,6 @@ import {
   Button,
   Card,
   CardActions,
-  CardActionsProps,
   CardContent,
   Container,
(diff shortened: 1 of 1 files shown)


amattu2 (approved):
Did not verify locally, but code changes LGTM!

PR #317, opened 2024-03-20
CRDCDH-881 Data Submission > Submitted Data tab

amattu2:
### Overview

This PR introduces a new Data Submissions Dashboard tab called Submitted Data. The new tab offers a node type filter and a dynamic table which displays the uploaded nodes and their properties–The columns change based on the selected node.

> [!WARNING]
> This branch is based off CRDCDH-882, which included newly added testing features.

- Add new tab
- Add test coverage for
  - General functionality (error handling)
  - Table filters
  - Table itself
  - Safe JSON parsing utility
- Remove unused "User" function argument on GenericTable > `renderValue`
- Added test ID attributes to GenericTable > Header Col and Pagination components

### Related Ticket(s)

CRDCDH-881


Alejandro-Vega (changes requested):
If possible could we move the overflow to within just the Table, without the table pagination section? Seems a bit inconvenient to have to scroll back to click on next page. Everything else LGTM.

amattu2:
> @Alejandro-Vega If possible could we move the overflow to within just the Table, without the table pagination section? Seems a bit inconvenient to have to scroll back to click on next page. Everything else LGTM.

Should be fixed now. Added a new prop to the table, `horizontalScroll` which allows the content to fill necessary space (disables line wraps) and adds a horizontal scrollbar to the table itself. The default is `false` and should have no effect on existing implementations.

Screenshot 2024-04-02 at 12 22 38 PM

Alejandro-Vega (changes requested):
One last thing I noticed was the table looks a little off when empty. I'm assuming it is due to there not being any columns. Not sure what the expected behavior is since you are getting them dynamically.

Screenshot 2024-04-02 at 12 32 27 PM

amattu2:
> One last thing I noticed was the table looks a little off when empty. I'm assuming it is due to there not being any columns. Not sure what the expected behavior is since you are getting them dynamically.
> 
> Screenshot 2024-04-02 at 12 32 27 PM

Byproduct of using `display: block` on a table. The `horizontalScroll` property is now ignored when there is no content. Thanks for finding this, I didn't even think to check that.

Screenshot 2024-04-02 at 12 53 16 PM

Alejandro-Vega (approved):
LGTM!

PR #323, opened 2024-03-25
CRDCDH-716 Added Submission Type/Intention when creating new Submission

Alejandro-Vega:
### Overview

Updated create a new data submission dialog to include submission type/intention radio input with "New", "Update", "Delete" options.

- Separated create new data submission dialog logic into separate component
- Made small adjustments to spacing and made close button more consistent with other dialogs
- Replaced deprecated Generic Alert with Notistack
- Added Radio input to the create dialog
- Updated Radio input icon sizing and styling
- Updated createSubmission mutation to include `intention` param
- Updated getSubmission query to return `intention` property
- Updated Submission type to include `intention`

### Related Ticket(s)

[CRDCDH-716](https://tracker.nci.nih.gov/browse/CRDCDH-716)

### DIFF
--- a/src/content/dataSubmissions/CreateDataSubmissionDialog.tsx
+++ b/src/content/dataSubmissions/CreateDataSubmissionDialog.tsx
@@ -1,0 +1,386 @@
+import { FC, useRef, useState } from "react";
+import { Box, Button, Dialog, DialogActions, DialogContent, DialogTitle, Grid, IconButton, Stack, Typography, styled } from "@mui/material";
+import { LoadingButton } from "@mui/lab";
+import { useMutation, useQuery } from "@apollo/client";
+import { query as approvedStudiesQuery, Response as approvedStudiesRespone } from "../../graphql/listApprovedStudiesOfMyOrganization";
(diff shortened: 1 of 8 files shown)


src/content/dataSubmissions/CreateDataSubmissionDialog.tsx
amattu2, line 1: Just a general suggestion for the create submission dialog form – since we're doing a light refactor of this component, now could be an opportunity to convert the form to [React Hook Form](https://react-hook-form.com/get-started#Registerfields) instead of managing the refs/states at the component level.

The user profile page is an example of how to use react-hook-form with TextBoxes and Selects:

https://github.com/CBIIT/crdc-datahub-ui/blob/3.0.0/src/content/users/ProfileView.tsx

This is completely optional and can be merged without the change.

amattu2 (changes requested):
Generally everything LGTM. The only issue is with the create dialog – since the content is longer now we need to make it scrollable.

amattu2 (approved):
LGTM! Nice job

PR #328, opened 2024-04-02
CRDCDH-974 Auto refresh Validation Results table on validate complete

amattu2:
### Overview

MVP-2.1.0 PR to auto refresh the Validation Results table upon starting or finishing the validation process. Also includes a bug fix for the Validation button itself, where it would never change from "Validating..." if the metadata/file validation status remained the same.

e.g. You validate one time and `metadataValidationStatus` is `Error`. You validate again, without uploading anything new, the status quickly becomes `Error` again and the button never sees any change from `Error` -> `Validating` -> `Error` so it just stays `Validating`

- Force immediate Data Submission refetch on validate (bug)
- Refresh QC Results table any time the validation properties update (linked ticket)

### Related Ticket(s)

CRDCDH-974

### DIFF
--- a/src/content/dataSubmissions/DataSubmission.tsx
+++ b/src/content/dataSubmissions/DataSubmission.tsx
@@ -481,6 +481,8 @@ const DataSubmission: FC<Props> = ({ submissionId, tab = URLTabs.DATA_ACTIVITY }
       return;
     }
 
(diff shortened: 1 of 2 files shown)


Alejandro-Vega (approved):
LGTM!

PR #339, opened 2024-04-08
CRDCDH-885 Pre-configured Uploader CLI Config

amattu2:
### Overview

This PR adds functionality to download a pre-configured Uploader CLI config file. 

- Rename original "DataUpload" to "MetadataUpload" (reflects what it actually does)
- Create "DataUpload" component which provides contextual info on how to upload data
  - Add test coverage for the requirements and general functionality (i.e. error handling)
- Create Uploader Config Dialog with the "data file and manifest path" form
  - Add basic test coverage for dialog functionality  
  - NOTE: The design indicates these fields are required, neither the ticket nor API indicate so
- Move Uploader CLI Dialog out of "user" content to a separate component
  - Add basic test coverage for dialog functionality
  - Fix an issue in which the tool download link would briefly open a new tab to download, now opens in current tab (K.C. requested)

### Related Ticket(s)

CRDCDH-885


src/components/DataSubmissions/DataUpload.tsx
```diff
@@ -1,386 +1,142 @@
...
-    alignItems: "center",
-    marginLeft: "20px",
+const StyledToolButton = styled(Typography)({
+  color: "#005999",
+  cursor: "pointer",
+  lineHeight: "16px",
+  fontWeight: 700,
+  display: "inline",
+  "& span": {
+    borderBottom: "1px solid #D1D1D1",
```
Alejandro-Vega, line 46: Small thing I noticed, underline and icon seems a bit off from the design. The icon is not a big deal so we can disregard that if you prefer. 

Figma:
Screenshot 2024-04-30 at 4 05 33 PM

Actual:
Screenshot 2024-04-30 at 4 08 10 PM

src/components/DataSubmissions/DataUpload.tsx
```diff
@@ -1,386 +1,142 @@
...
+const StyledDownloadButton = styled(Button)({
+  padding: "10px 6px",
+  fontSize: "14px",
   fontStyle: "normal",
-  lineHeight: "24px",
+  lineHeight: "16px",
   letterSpacing: "0.32px",
-  textTransform: "none",
+  width: "137px",
+  height: "44px",
```
Alejandro-Vega, line 33: Other buttons are 46px, we should keep this consistent.

Alejandro-Vega (changes requested):
Nice job! Looks good, just pointed out some small styling issues.

src/components/DataSubmissions/DataUpload.tsx
amattu2: Good catch. I fixed the underline, it should be much closer to the Figma styling now.  a6b7266

I left the icon as the one from MUI since the difference is relatively minor and it's one less asset to manage.

src/components/DataSubmissions/DataUpload.tsx
amattu2: Good catch. It seems the buttons have a calculated height of 47px, so I updated the styling to that.  a6b7266

For this button, we can't omit the height attribute, otherwise it grows too big (multiple text lines).

Alejandro-Vega (approved):

PR #348, opened 2024-04-25
Hotfix: Added uploadType and dataSubmission to deps array for validate button

Alejandro-Vega:
### Overview

Added additional `uploadType` and `dataSubmission` to deps of the memoized validate button. Please double check that I didn't miss any additional missing deps.

N/A

### Related Ticket(s)

N/A

### DIFF
--- a/src/components/DataSubmissions/ValidationControls.tsx
+++ b/src/components/DataSubmissions/ValidationControls.tsx
@@ -173,8 +173,7 @@ const ValidationControls: FC<Props> = ({ dataSubmission, onValidate }: Props) =>
       </StyledValidateButton>
     ),
     [
-      validationType,
-      uploadType,
+      handleValidateFiles,
       dataSubmission,
       canValidateFiles,
       canValidateMetadata,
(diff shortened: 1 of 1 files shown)


src/components/DataSubmissions/ValidationControls.tsx
```diff
@@ -172,7 +172,15 @@ const ValidationControls: FC<Props> = ({ dataSubmission, onValidate }: Props) =>
         {isValidating ? "Validating..." : "Validate"}
       </StyledValidateButton>
     ),
-    [validationType, canValidateFiles, canValidateMetadata, isValidating, isLoading]
+    [
+      validationType,
+      uploadType,
```
amattu2, line 177: This should work, but I'm wondering if we should just simplify it by adding `handleValidateFiles` as a dependency instead of `validationType` and `uploadType`? 

If `handleValidateFiles` has additional dependencies in the future, we wouldn't need to add them in here.

amattu2 (approved):
We can merge as-is, or with the alternative approach. Either way is fine with me.

src/components/DataSubmissions/ValidationControls.tsx
Alejandro-Vega: Good point! I like your suggestion better

PR #355, opened 2024-05-01
CRDCDH-1080 PMVP Updated Submission Name input validation to trim whitespace

Alejandro-Vega:
### Overview

Updated the validation for the "Submission Name" input field to trim the whitespace first.

### Change Details (Specifics)

> [!Note]
> This is the PMVP implementation for this fix. This required a different implementation due to refactoring in `3.0.0` branch.
> MVP2-M3 fix implemented in #354

### Related Ticket(s)

[CRDCDH-1080](https://tracker.nci.nih.gov/browse/CRDCDH-1080)

PR #357, opened 2024-05-03
fix: Submission Name validation not allowing submit

amattu2:
### Overview

This PR fixes an issue with creating a submission, where the createSubmission API is never called. 

To verify:

- Fill out the form WITHOUT causing any validation issues
- it should now send the API request

I wasn't able to replicate this when originally testing #355, so it's possible there's an issue beyond this – But this solution is working locally.

- Return `null` from the `validateEmpty` function if there's no issue

### Related Ticket(s)

N/A


Alejandro-Vega (approved):
Thanks for fixing this for me! Nice catch.

PR #362, opened 2024-05-07
CRDCDH-1027 Delete orphan files

Alejandro-Vega:
### Overview

Added ability to delete individual or all orphan files within Validation Results tab. Also, added an additional table pagination to the top of the table.

- Added individual deletion of orphan files within the "Issues" column of the Validation Results table. When clicked, it will delete the orphan file and refresh the table. This uses the `Submission.fileErrors` to determine orphan files
- Added bulk/all deletion of orphan files within the "Issues" column of the Validation Results table. When clicked, it will delete all orphan files and refresh the table
- Added tooltip to the delete all orphan files icon button
- Updated Validation Results table to have a second pagination section above the table headers
- Updated filter styling in Validation Results tab and migrated to use the form components select to be consistent

**Testing Steps:**
- Upload Data Files via CLI
- Validate
- Navigate to Validation Results
- Observe orphaned file(s) buttons, tooltip, as well as top table pagination. (Delete all orphaned files button will open a dialog)
- After all orphaned files are deleted, it requires an additional validation to resolve the "Error" state for `fileValidationStatus`.

### Related Ticket(s)

[CRDCDH-1027](https://tracker.nci.nih.gov/browse/CRDCDH-1027)


src/components/DataSubmissions/DeleteOrphanFile.tsx
amattu2, line 1: Is this component used somewhere? If so, should it say "Delete extra file" or "Delete orphaned file"?

src/components/DataSubmissions/DeleteAllOrphanFilesButton.tsx
amattu2, line 1: Question on this requirement:

> These delete buttons should only be present when there are extra file errors displayed on the Validation Results tab.

The current implementation for the Delete All button is consistent with the export button, which is always present but disabled. 

Do we know which is correct?

src/components/DataSubmissions/DeleteAllOrphanFilesButton.test.tsx
```diff
@@ -0,0 +1,358 @@
...
+    await waitFor(() => {
+      expect(screen.getByText("Delete All Orphaned Files")).toBeInTheDocument();
+    });
+
+    userEvent.click(getByTestId("delete-dialog-confirm-button"));
+
+    await waitFor(() => {
+      expect(onDelete).toHaveBeenCalledWith(true);
+    });
+  });
```
amattu2, line 200: I think this case is testing the same thing as below? The only difference I see is the missing alert assertion in this one. No action necessarily needed here, just wanted to point it out in case it's an accident.

src/components/DataSubmissions/DeleteOrphanFileChip.tsx
```diff
@@ -0,0 +1,135 @@
...
+      const { data: d, errors } = await deleteOrphanedFile({
+        variables: {
+          _id: submission._id,
+          fileName: submittedID,
+        },
+      });
+
+      if (errors || !d?.deleteOrphanedFile?.success) {
+        throw new Error("Unable to delete orphan file.");
+      }
```
amattu2, line 101: Since we don't enforce coverage on CRDC-DH, this is **COMPLETELY** optional–but this branch doesn't seem to be covered by the tests.

Screenshot 2024-05-16 at 2 38 21 PM

I don't know if it's the missing scenario for `success: false` or maybe missing GraphQL error scenario since Apollo has two different types of errors (network vs GraphQL level). e.g.

```javascript
  {
    request: {
      query: DELETE_ORPHANED_FILES,
      variables: { _id: "submission-id" },
    },
    result: {
      errors: [new GraphQLError("Mock GraphQL error")],
    },
  },
```

src/components/DataSubmissions/DeleteAllOrphanFilesButton.tsx
```diff
@@ -0,0 +1,133 @@
...
+    try {
+      const { data: d, errors } = await deleteAllOrphanedFiles({
+        variables: {
+          _id: submission._id,
+        },
+      });
+
+      if (errors || !d?.deleteAllOrphanedFiles?.success) {
+        throw new Error("Unable to delete all orphaned files.");
+      }
```
amattu2, line 83: Again, **completely optional**. This branch isn't covered either. 

Screenshot 2024-05-16 at 2 38 09 PM

amattu2 (changes requested):
Really solid job and awesome to see component-level test coverage.

Few things annotated below, some of which can be ignored.

src/components/DataSubmissions/DeleteAllOrphanFilesButton.tsx
Alejandro-Vega: I confirmed previously with Kailing that they should always be shown but disabled to be consistent. Although, it seems like the requirements were never updated. Left a comment on the ticket.

src/components/DataSubmissions/DeleteOrphanFile.tsx
Alejandro-Vega: Removed unused component, I replaced it with a different component and forgot to delete it.

src/components/DataSubmissions/DeleteAllOrphanFilesButton.test.tsx
Alejandro-Vega: Removed test since it is already covered by another test. Thanks!

src/components/DataSubmissions/DeleteOrphanFileChip.tsx
Alejandro-Vega: Added both, but it seemed to be caused by the success false condition.

src/components/DataSubmissions/DeleteAllOrphanFilesButton.tsx
Alejandro-Vega: Added coverage.

amattu2 (approved):
LGTM!

PR #367, opened 2024-05-14
CRDCDH-1125 Fix Active Submissions Dialog

amattu2:
### Overview

This PR addresses an issue with the active data submissions dialog that appears when removing a study from an organization. Some refactoring in M3 caused the logic to break, always showing the confirmation dialog, even when no studies were removed. 

- Map the form selectbox input from a Study ID back to a Study Abbreviation before comparing whether or not a study was removed

### Related Ticket(s)

CRDCDH-1125 (Bug)
CRDCDH-860 (Broke the feature)
CRDCDH-689 (Original Feature)


Alejandro-Vega (approved):
LGTM!

PR #368, opened 2024-05-14
CRDCDH-1137 Allow validation in Submitted state

amattu2:
### Overview

This PR covers a missing requirement that Data Curators or Admins should be able to validate a Submitted data submission. These changes should only effect Data Curators and Admins for a Data Submission in the "Submitted" state – Nothing else should change...

> [!NOTE]
> When migrating to v3.0.0, I have test coverage for these changes written already, I just need to migrate them.

- Rewrite the validate permission map to be a simplified object.
  - `Key`: Submission Status
  - `Value`: List of Roles that can validate
- Create a util to determine what the default Validation Target should be, we had one for the Validation Type already
- Disable "New" validation target if the submission is Submitted
  - This is already disabled in statuses beyond "Submitted"

### Related Ticket(s)

CRDCDH-1137


Alejandro-Vega (approved):
Logic looks good to me and is working as expected! Nice job!

PR #371, opened 2024-05-16
CRDCDH-1138 Reduce Data Submission polling from 60s to 1s

Alejandro-Vega:
### Overview

Reduced the getSubmission and listBatches polling from 60s to 1s intervals. Also fixed race-condition bug involving Validate button and polling.

- Added polling for listBatches API
- Set polling interval for getSubmission and listBatches to 1s
- Memoized Data Submission components as the frequent 1s polling would cause excessive re-renders. 
- Fixed bug where clicking the validate button would set the button text to "Validating..." and would infinitely poll getSubmission even though it is done validating.
  
  This was due to a race condition where if the `metadataValidationStatus` is set to "Error" and when an additional validate is started with no changes it would poll getSubmission, but receive the same data back with `metadataValidationStatus` still as "Error". Receiving the same data back made it so the component wouldn't re-render and check the validation status again to stop the polling and update the button text. Added key that updates on every poll attempt to fix issue.

### Related Ticket(s)

[CRDCDH-1138](https://tracker.nci.nih.gov/browse/CRDCDH-1138)

### DIFF
--- a/src/components/DataSubmissions/BackButton.tsx
+++ b/src/components/DataSubmissions/BackButton.tsx
@@ -1,4 +1,4 @@
-import { FC } from 'react';
+import React, { FC } from 'react';
 import { useNavigate } from 'react-router-dom';
 import ArrowBackIcon from '@mui/icons-material/ArrowBack';
 import { Button, ButtonProps, styled } from '@mui/material';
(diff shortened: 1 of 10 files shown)


amattu2 (approved):
Everything LGTM and seems to work as expected. Awesome job with the memoization QoL updates.

PR #375, opened 2024-05-20
CRDCDH-1059 Add "404 – Not found" page

amattu2:
### Overview

This PR addresses two issues:

- We previously had a placeholder 404 page with just some plain text
- The 404 page would not render in all scenarios due to a home route catch-all

- Implement 404 page design
  - Added test coverage for this page and one snapshot test
- Remove unused home (`/`) query catch-all
- Fix missing <RequireAuth> on Questionnaire view (it was still protected by higher-up routes)

### Related Ticket(s)

CRDCDH-1059 (bug/feature)
CRDCDH-1077 (design)


Alejandro-Vega (approved):
LGTM! Works as expected

PR #376, opened 2024-05-20
Fix: Default CRA Eslint Config Disabled

amattu2:
### Overview

At some point in development (likely early on), we added the `extends` option to our Eslint config. In doing so, we overrode the base `react-app` and `react-app/jest` configurations originally defined in our `package.json` file. This PR primarily addresses that oversight by moving those two base configs into our `.estlintrc.cjs` file. 

At some point in the future, we would benefit from enabling these options as missing deps has caused bugs:

- react-hooks/rules-of-hooks
- react-hooks/exhaustive-deps

These were disabled due to the amount of code-changes needed, and the potential for unknown side-effects of applying a fix without comprehensive testing.

- Move the `package.json` `eslintConfig` option to the `.eslintrc.cjs` file
- Disable these 7 newly added rules provided by CRA:
  - rules-of-hooks
  - exhaustive-deps
  - prefer-screen-queries
  - no-wait-for-multiple-assertions
  - no-container
- Add global definition for `window` and `document` (eslint)
- Enforce styleguide on public/js files
- Update browser-list
- Update default package.json pkg version to reflect `3.0.0`
- Update default `manifest.json` to reflect CRDC verbiage

### Related Ticket(s)

N/A


Alejandro-Vega (approved):
LGTM! These are some great changes, thanks for adding these!

PR #379, opened 2024-05-22
CRDCDH-1181 Adjust Individual Node Chart Carousel

amattu2:
### Overview

This PR adjusts the individual node chart display to show the 3rd node as half-faded for large datasets. For datasets under 4 nodes, it will show up to 3 without any fading.

<details><summary>Screenshots</summary>
<p>
**One Node**
1 Node

**Two Nodes**
2 Nodes

**Three Nodes**
3 Nodes

**Four Nodes**
4 Nodes

**9 Nodes (same as anything above 5)**
9 Nodes
</p>
</details> 

- Add additional `locked` prop to the carousel component which disables fading and user interaction
- Conditionally apply the left-side fading based on current carousel index
- Add a buffer element to the carousel when there's 4 nodes to prevent the 4th node from being cut in half

### Related Ticket(s)

CRDCDH-1181

### DIFF
--- a/src/components/Carousel/index.tsx
+++ b/src/components/Carousel/index.tsx
@@ -1,4 +1,4 @@
-import { FC } from 'react';
+import { FC, useState } from 'react';
 import { styled } from '@mui/material';
 import Carousel, { CarouselProps } from 'react-multi-carousel';
 import 'react-multi-carousel/lib/styles.css';
@@ -7,6 +7,10 @@ import CustomRightArrow from './CustomRightArrow';
 
 type Props = {
   children: React.ReactNode;
+  /**
+   * If true, will disable any user interaction with the carousel.
+   */
+  locked?: boolean;
 } & Partial<CarouselProps>;
 
 const sizing = {
@@ -16,7 +20,9 @@ const sizing = {
   },
 };
 
-const StyledWrapper = styled('div')({
+const StyledWrapper = styled('div', {
+  shouldForwardProp: (p) => p !== "showLeftFade" && p !== "showRightFade"
+})<{ showLeftFade: boolean; showRightFade: boolean }>(({ showLeftFade, showRightFade }) => ({
   maxWidth: "700px",
   minWidth: "464px", // NOTE: Without a min-width, the carousel collapses to 0px wide
   width: "100%",
@@ -37,7 +43,7 @@ const StyledWrapper = styled('div')({
     left: "calc(100% - 162px)",
     top: "0",
     bottom: "0",
-    width: "162px",
+    width: showRightFade ? "162px" : "0px",
     background: "linear-gradient(to right, rgba(255,255,255,0) 0%, rgba(255,255,255,1) 50%)",
     zIndex: 9,
   },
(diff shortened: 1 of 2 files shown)


Alejandro-Vega (changes requested):
Everything looks good to me, but I have a question regarding the carousel items sliding. It seems that every time I slide the carousel using the arrows, the donuts are being slightly offset more and more. Is this something that can be easily fixed? Or should we disregard since it is so minor and not likely to come up? If we want to disregard, I will just merge it in. Here's some examples:

Screenshot 2024-05-23 at 11 24 11 AM
Screenshot 2024-05-23 at 11 24 38 AM
Screenshot 2024-05-23 at 11 28 39 AM

Eventually, if you have enough nodes (although not likely at all) the last slide looks a bit off. Here's what the last slide looks like after 12 nodes. (To test this I spread the dataset into an array a few times `[...dataset, ...dataset, ...dataset]?.map( ...`)
Screenshot 2024-05-23 at 11 32 03 AM

amattu2:
> @Alejandro-Vega  Everything looks good to me, but I have a question regarding the carousel items sliding. It seems that every time I slide the carousel using the arrows, the donuts are being slightly offset more and more. Is this something that can be easily fixed? Or should we disregard since it is so minor and not likely to come up? If we want to disregard, I will just merge it in. 

Yeah I think this is something we should fix. It's always been somewhat offset, so I'm looking into what I can do to address this.

amattu2:
@Alejandro-Vega A partial fix for the large offset was applied in 34c21e7. It still isn't perfectly centered when using the arrows, and somewhat aggressively scrolls for submissions with only 4 nodes. For large submissions (10+ nodes), the giant offset at the end should now be gone. 

If this is WORSE than before, LMK and I'll just revert 34c21e7. Otherwise, this might be as good as I can fine-tune the library ATM.

Alejandro-Vega:
> @Alejandro-Vega A partial fix for the large offset was applied in [34c21e7](https://github.com/CBIIT/crdc-datahub-ui/commit/34c21e791f1d31bae44f02fae16f0c955e8210db). It still isn't perfectly centered when using the arrows, and somewhat aggressively scrolls for submissions with only 4 nodes. For large submissions (10+ nodes), the giant offset at the end should now be gone.
> 
> If this is WORSE than before, LMK and I'll just revert [34c21e7](https://github.com/CBIIT/crdc-datahub-ui/commit/34c21e791f1d31bae44f02fae16f0c955e8210db). Otherwise, this might be as good as I can fine-tune the library ATM.

It seems to now skip some nodes around the 10th node which translates the carousel from -1550px to -1850px, meanwhile all the others translate by -200px. I tried to debug this myself, but not sure why this is occurring. This might require a deeper look into the package itself. I suggest we revert this change and leave it as it was since this is not a huge deal. We can revisit this further at a future date.

amattu2:
> > @Alejandro-Vega A partial fix for the large offset was applied in [34c21e7](https://github.com/CBIIT/crdc-datahub-ui/commit/34c21e791f1d31bae44f02fae16f0c955e8210db). It still isn't perfectly centered when using the arrows, and somewhat aggressively scrolls for submissions with only 4 nodes. For large submissions (10+ nodes), the giant offset at the end should now be gone.
> > If this is WORSE than before, LMK and I'll just revert [34c21e7](https://github.com/CBIIT/crdc-datahub-ui/commit/34c21e791f1d31bae44f02fae16f0c955e8210db). Otherwise, this might be as good as I can fine-tune the library ATM.
> 
> It seems to now skip some nodes around the 10th node which translates the carousel from -1550px to -1850px, meanwhile all the others translate by -200px. I tried to debug this myself, but not sure why this is occurring. This might require a deeper look into the package itself. I suggest we revert this change and leave it as it was since this is not a huge deal. We can revisit this further at a future date.

Reverted the commit in 4af3775.

PR #383, opened 2024-05-24
CRDCDH-1094 Create new Data Submission changes. New intention and added dataType

Alejandro-Vega:
### Overview

Updated the create new data submission dialog to consider new Submission intention as well as adding the Submission Data Type property.

- Moved intention radio options in create submission dialog to the top of the dialog instead of below.
- Added `dataType` property to Submission. This will be for specifying if the submission is for both metadata and data files, or just metadata.
- Updated getSubmission query and createSubmission mutation to include `dataType`
- Updated type of `intention` property to contain "New/Update" instead of having them separate options
- Updated tests due to the new type changes as well as the addition of property `dataType`
- Disabled "Metadata and Data Files" dataType option when intention is "Delete"
- The dataType option "Metadata and Data Files" is set to selected by default

### Related Ticket(s)

CRDCDH-1094


amattu2 (approved):
LGTM!

PR #398, opened 2024-06-06
Fix silent warnings allowing Jest to succeed

amattu2:
### Overview

This PR prevents console.warn and console.error calls from silently happening in tests and not failing the Jest action for it.

- Add `jest-fail-on-console` dependency

### Related Ticket(s)

N/A


Alejandro-Vega (approved):
LGTM!

PR #409, opened 2024-06-25
CRDCDH-1183 Migrate Data Submission List table to GenericTable and add search params

Alejandro-Vega:
### Overview

Migrated Data Submission List table to use GenericTable component instead, with search params in URL enabled.

- Migrated Data Submission List table to use GenericTable component instead, with search params in URL enabled
- Updated filters to use react-hook-form
- Added search params support for filters above table
- Updated styling around filters to better match design
- Updated back button text in Data Submission Dashboard
- Updated Site Navigation to still show underline under Data Submission when search params are applied to the table

### Related Ticket(s)

[CRDCDH-1183](https://tracker.nci.nih.gov/browse/CRDCDH-1183)


src/content/dataSubmissions/DataSubmissionsListView.tsx
```diff
@@ -245,76 +217,163 @@ const ListingView: FC = () => {
...
+    if (!activeOrganizations?.length) {
+      return;
+    }
+
+    try {
+      setLoading(true);
+
+      const { data: d } = await refetch();
+      if (error || !d?.listSubmissions) {
+        throw new Error("Unable to retrieve submission quality control results.");
```
amattu2, line 359: Could we update this error message to something else? I don't believe it's actually used anywhere, but if it ever is, this would be an issue.

amattu2 (changes requested):
One minor QoL change annotated. Everything else LGTM.

src/content/dataSubmissions/DataSubmissionsListView.tsx
Alejandro-Vega: Oops copy paste error. Thanks!

amattu2 (approved):
Didn't run it again locally after the change, but LGTM.

PR #403, opened 2024-06-12
CRDCDH-1155 Data Submission Validation Status

amattu2:
### Overview

This is a large PR to introduce Data Submission refactoring and a new "Validation Status" badge, among other things.

> [!NOTE]
>   Going forward, new (or refactored) components should rely on the `useSubmissionContext` hook instead of prop-drilling. I only updated the components related to this feature but left everything else alone.

### Change Details (Specifics)

- Migrate the Data Submission `getSubmission` query to a context provider
  - NOTE – Abstracting the query is required so we can make the native Apollo methods available to deeply nested components without prop-drilling
- Migrate ValidationControls to the new Data Submission Context and update tests for it
- Remove forced re-rendering of ValidationControls
  - NOTE – This is no longer necessary due to how we handle the state updates and watching the current validation status
- Remove the automatic reset of the current Validation Target/Type when initiating validation, and instead, reset AFTER validation ends
  - NOTE – This does not persist through a page reload. Only if you stay on the page
- Add a "Input" type definition for the `validateSubmission` query
- Fix an inconsistency with the ValidationType type definition
  - Remove `All` (metadata and files) as this is a frontend-only concept
  - Add missing `cross-submission`
  - `Metadata` -> `metadata`
  - `Files` -> `file`

### Related Ticket(s)

CRDCDH-1155

PR #412, opened 2024-06-27
CRDCDH-1240 Inconsistent Submission/Batch Polling

amattu2:
### Overview

This PR builds ontop of #403 to add automatic polling at the `getSubmission` level if there's a uploading batch. This PR also separates the Data Activity tab from the broader Data Submission page, which addresses another issue where the batch list would not reflect any new batches created when uploading from a tab other than Data Activity.

> [!NOTE]
> The DataActivity component does not have full test coverage and a few tests are marked as TODOs for the future.
The tests were failing due to their implementation (not code issues) and because of time constraints, I decided to skip fixing them for now.

> [!NOTE]
> Some (unknown) change in this PR forced the Statistic Carousel to not rerender as often, which causes the navigational arrows to not appear when more than 3 donut charts are present. To bypass this, I added a key to the carousel element which is dynamic based on the node count.

> [!Warning]
> One of the tests in FileListDialog is disabled because it is (correctly) failing. We will eventually need to address this issue, but it's out of scope for now.

(Ordered from most relevant to least)

- Separate Batch Table into a "DataActivity" component
  - FYI – this will be renamed in Sprint 6
  - Add basic test coverage, see note above about incomplete tests
  - This component will start `getSubmission` polling if it finds a uploading batch and the submission object is NOT polling already
- Add batch status to `getSubmission` query and auto poll if a batch is uploading
  - Update SubmissionCtx coverage to ensure it polls when there are uploading batches
- Add start/stop polling wrappers to the SubmissionCtx so we can internally track if polling is ongoing
  - Apollo doesn't expose any flag to indicate if you have started polling a query, so this is the work around
- Move the ErrorDialog (dialog with the list of validation/batch errors) to a separate components folder
  - Add test coverage
- Move the FileListDialog (batch files list) to a separate components folder
  - Add basic test coverage, but not 100%
- Force the Carousel to rerender on statistic length change

### Related Ticket(s)

CRDCDH-1240


Alejandro-Vega (approved):
LGTM!

PR #423, opened 2024-07-08
CRDCDH-1213 Submitted Data > Support Data Deletion 

amattu2:
### Overview

This PR introduces the ability to delete submitted data (metadata and data files) from within the Data View tab, among other misc supporting changes. In general, the deletion functionality revolves around selecting rows from the table, and the selection should persist between pagination. If deletion is ongoing, the submission will poll until it's completed. These changes also include additional filtering for the Submitted Data tab. 

> [!NOTE]
> When deleting all from a specific node type, the filter box is cleared and the table empties out. The requirements don't say to select the next node in the dropdown, so I left this as-is, but this may need to be updated in the future.

> [!Warning]
> There are a few important QoL action items that need resolved:
> - The SubmissionContext is forcing almost all of the components to rerender due to state changes when polling. This is caused by the `status` attribute switching between LOADING and LOADED, as well as the submissionStats order changing (the API does not sort the array and it's random)
> - The Data View table is fetching the data twice when switching nodes because we don't have the proper `orderBy` attribute specified upon loading. 

(Ordered by significance/relevance to the story)

- Create `DeleteNodeDataButton` to handle node deletion and confirmation
  - Create test coverage for this component
- Implement DeleteNodeDataButton in SubmittedData along with checkboxes in the table
  - Added/changed coverage for the new requirements in this tab
  - Also migrate SubmittedData from prop-drilling to the SubmissionContext
- Add "Node Status" and "Submitted ID" to the Submitted Data Filters
  - Implement coverage for filter requirements (e.g. debounced queries)
  - Minor restyling to match Quality Control filter design
- Add Data Submission context automatic polling if deleting data
  - Added coverage for this
- Add `deletingData` property to submission (affects many test files too)
- Rename visual references of "Submitted Data" to "Data View"
- Migrate Data Submission `DeleteDialog` to a separate components file
  - Create basic test coverage for the component
- Fixed an issue with GenericTable using `[Object object]` for the `<th>` key and testid when the `label` is a element
  - Added initial test coverage for GenericTable "test driven development" support

### Related Ticket(s)

CRDCDH-1213


src/content/dataSubmissions/SubmittedData.tsx
```diff
@@ -152,20 +230,27 @@ const SubmittedData: FC<Props> = ({ submissionId, submissionName }) => {
       return;
     }
 
-    if (error || !d?.getSubmissionNodes || !d?.getSubmissionNodes?.properties?.length) {
+    if (error || !d?.getSubmissionNodes || !("properties" in d.getSubmissionNodes)) {
```
Alejandro-Vega, line 233: Optional alternative, this is the same as `d.getSubmissionNodes?.includes("properties")`

src/content/dataSubmissions/SubmittedData.tsx
```diff
@@ -177,53 +262,159 @@ const SubmittedData: FC<Props> = ({ submissionId, submissionName }) => {
...
+      setSelectedItems(data?.map((node) => node.nodeID) || []);
+    });
+
+    // If all rows are already visible, no need to fetch data
+    if (data?.length === totalData) {
+      isFetchingAllData.current = false;
+      return;
+    }
+
+    const { data: d, error } = await getSubmissionNodes({
```
Alejandro-Vega, line 303: This should be wrapped in a try/catch otherwise will cause uncaught exception.

Alejandro-Vega (changes requested):
Super solid job! I only request one small change, as well as one optional but it is preference. Everything is looks great to me! Also seems to work as expected. Awesome work.

src/content/dataSubmissions/SubmittedData.tsx
amattu2: Thanks for catching this! Fixed in 8564282

Alejandro-Vega (approved):

PR #430, opened 2024-07-17
update DMN package

ty-esi:
### Overview

Update DMN package

- update DMN package.

### Related Ticket(s)

- https://github.com/CBIIT/Data-Model-Navigator/pull/148
- https://github.com/CBIIT/Data-Model-Navigator/pull/147
- https://github.com/CBIIT/Data-Model-Navigator/pull/146


src/config/DataCommons.ts
amattu2, line 1: When exporting the individual file node for ICDC, it's using the name `ICDC__v1.0.0`. Are you able to replicate this issue? I think we'll also need to specify the fileTransferManifestName for ICDC as well.

src/config/DataCommons.ts
```diff
@@ -27,7 +27,7 @@ export const DataCommons: DataCommon[] = [
         fileType: "pdf",
         prefix: "CDS_",
         downloadPrefix: "CDS_",
-        fileTransferManifestName: "CDS_Data_Loading_Template-file-manifest",
+        fileTransferManifestName: "CDS_Data_Loading_Template_File_Manifest",
```
amattu2, line 30: My assumption with the bug ticket CRDCDH-1360 would be that we want to be completely consistent with the rest of the node TSV names. If that's the case, the correct values would be:

|Data Common|fileTransferManifestName|Note|
|:-|:-|:-|
|CDS| `CDS_Data_Loading_Template_file`|
|CTDC|`CTDC_Data_Loading_Template_data_file`|CTDC uses "data_file" instead
|ICDC|`ICDC_Data_Loading_Template_file `|

Any thoughts on this? I might be misinterpreting the bug

PR #433, opened 2024-07-22
CRDCDH-1214 Cross Validation Tab

amattu2:
### Overview

This PR re-introduces the ability to view Cross Validation issues, but it's now under a separate tab called Cross Validation Results. This tab is only visible to Admins/Curators and after cross validation was started.

- Create a new Cross Validation tab
  - Create test coverage for requirements/general functionality
- Create a component to handle the cross validation filters
  - Add test coverage for expected functionality
- Make minor updates to existing QualityControl tab
  - Remove unused `refreshSubmission` prop
  - Remove forced table-refresh when `crossSubmissionStatus` changes (now in a separate tab)
  - Update the table's "no content" message per requirement 5
- Centralize the "CrossValidateRoles" variable instead of duplicating in multiple components
- Merge the redundant `DataValidationResult` type into the existing `AsyncProcessResult` type
- Fix an issue with the apollo cache of the batch list causing refetching of misc. queries

### Related Ticket(s)

CRDCDH-1214


src/content/dataSubmissions/CrossValidation.tsx
```diff
@@ -0,0 +1,304 @@
...
+    renderValue: ({ conflictingSubmission: _id }) => (
+      <span key={_id} data-testid={`conflicting-submission-${_id}`}>
+        <StyledFormTooltip title={_id}>
+          <Link
+            to={`/data-submission/${_id}`}
+            target="_blank"
+            data-testid={`conflicting-link-${_id}`}
+          >
+            ...
+            {_id?.slice(-5)}
```
Alejandro-Vega, line 91: I see that an example resulting value is displayed as `...c54"]`. I wonder if the quotes and square bracket count towards the "last 5 digits" displayed as well, or if they meant just the id like: `...7fc54`. If your original way is correct then disregard this comment.

Alejandro-Vega (changes requested):
Everything looks solid to me. Great work! Added one optional comment regarding the requirements. Feel free to disregard if it looks good the way it is.

src/content/dataSubmissions/CrossValidation.tsx
amattu2: Fixed in the submission `cfa56943-6e0f-4c9a-bac7-1b7c90313a2d`. This was a data issue due to how conflicting IDs used to be stored as a JSON string instead of an array.

PR #434, opened 2024-07-24
CRDCDH-1334 Disable Create Submission button when user has Inactive org

Alejandro-Vega:
### Overview

Disabled "Create a Data Submission" button in the Data Submission List page when the user belongs to an "Inactive" organization. I also added a tooltip to the button for when it is disabled.

- Disabled "Create a Data Submission" button in the Data Submission List page when the user belongs to an "Inactive" organization
- Added tooltip to button to display a message when it is disabled and hovered
- Fixed listOrganizations return type
- Moved the dialog component into the "components" folder
- Added tests for the dialog, including checking for disabled button and tooltip showing when hovering over button when belonging to an inactive org
- Updated getMyUser query to include status with the organization

### Related Ticket(s)

[CRDCDH-1334](https://tracker.nci.nih.gov/browse/CRDCDH-1334)


src/components/DataSubmissions/CreateDataSubmissionDialog.tsx
```diff
@@ -550,14 +571,30 @@ const CreateDataSubmissionDialog: FC<Props> = ({ onCreate }) => {
...
+            <span>
+              <StyledButton
+                type="button"
+                variant="contained"
+                onClick={handleOpenDialog}
+                disabled={
+                  !hasOrganizationAssigned ||
+                  userHasInactiveOrg ||
+                  authStatus === AuthStatus.LOADING ||
+                  orgStatus === OrgStatus.LOADING
```
amattu2, line 590: Do we still need to check the Organization Context status? When navigating to the data submission page, it briefly renders as disabled because the organization list is still loading. 

```suggestion
                  authStatus === AuthStatus.LOADING
```

If it's needed still, you can dismiss this.

amattu2 (approved):
Nice job and thanks for adding coverage to the dialog, it was very much overdue for that.

I annotated one potential UX improvement (depending on if it's intentional or not). Approving as-is.

src/components/DataSubmissions/CreateDataSubmissionDialog.tsx
Alejandro-Vega: Good catch! Was there from before the BE was ready. Replaced the orgStatus check to instead check when approvedStudies data is loading, otherwise tests will fail.

amattu2 (approved):

PR #441, opened 2024-08-12
CRDCDH-1437 Fix Primary Chart Label Overlapping

amattu2:
### Overview

This PR re-writes the Data Submission primary chart X-axis label implementation to cover scenarios in which the node labels would overlap. It now checks the approximate width (pixels) of the label text, and rotate the label 65 degrees in the event that it would extend beyond the width of the bar that the label sits under (i.e. it would overlap with another label...). When the label is rotated and trimmed, a tooltip will now appear when hovering over it.

<details><summary>Before vs. After</summary>
<p>
Screenshot 2024-08-12 at 12 16 45 PM
</p>
<p>
Screenshot 2024-08-13 at 4 34 24 PM
</p>
</details> 

- Compute the X-Axis label width and rotate the entire set of labels 65deg if there would be any overlapping
- When rotating the labels, trim any excess characters that would cover the controls below it (see screenshot)
- Add a custom tooltip when hovering the X-Axis labels only if they are rotated
- Add `label` attribute to `BarChartDataset` type definition (it was always there, but never caused type errors...)
- Add `jest-canvas-mock` to prevent misc. test failures
- Add minor misc. test coverage for changed features

### Related Ticket(s)

CRDCDH-1437


src/components/NodeTotalChart/CustomTick.tsx
```diff
@@ -10,20 +13,63 @@ const StyledLabel = styled("tspan")({
...
-    <g transform={`translate(${x},${y})`}>
-      <text x={0} y={0} dy={6}>
-        {textLines.map((tspan, index) => (
-          <StyledLabel key={tspan} textAnchor="middle" x={0} dy={6 * index + 8}>
-            {tspan}
-          </StyledLabel>
-        ))}
-      </text>
+    <g transform={`translate(${x},${y})`} onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
+      <g transform={angled ? "rotate(65), translate(-4, 7)" : undefined}>
```
Alejandro-Vega, line 56: Is there a purpose for the translate? If so, we can probably disregard. But in extreme examples, in this case 50 bar charts, this does cause an misalignment. See screenshots:

<details><summary>With translate</summary>

image
</details>

<details><summary>Without translate</summary>

Screenshot 2024-08-15 at 12 56 12 PM
</details>

src/utils/statisticUtils.ts
```diff
@@ -95,3 +95,29 @@ export const calculateMaxDomain = (dataMax: number): number => {
...
+/**
+ * A utility function to compute the approximate width of a text element
+ * rendered on an SVG canvas.
+ *
+ * @see https://stackoverflow.com/questions/118241/calculate-text-width-with-javascript/21015393#21015393
+ * @param text The text to measure
+ * @param font The font descriptor for the text element
+ * @returns The computed width of the text element or 0 if the width is not available
+ */
+export const calculateTextWidth = (text, font = "normal 11px Arial") => {
```
Alejandro-Vega, line 108: Could we take the font weight, font size, and font family as separate params? Having a single param makes it easy to mess it up and it won't be immediately clear if it is used wrong.

src/components/NodeTotalChart/index.tsx
```diff
@@ -29,47 +30,98 @@ const StyledChartContainer = styled(Box)({
...
-          interval={0}
-          allowDataOverflow
-          allowDuplicatedCategory
+const NodeTotalChart: FC<Props> = ({ data, normalize = true }) => {
+  const [tooltipData, setTooltipData] = useState<{ label: string; x: number; y: number } | null>(
+    null
+  );
+
+  const computedBarWidth = useMemo<number>(
+    () => (482 - 8 * data.length) / data.length,
```
Alejandro-Vega, line 39: Could we assign variables to these numbers? Will make the logic easier to read.

Alejandro-Vega (changes requested):
Great solution with the canvas! It is very interesting and I have never considered using it this way. I tested the bar charts and was unable to cause any issues without using extreme/unlikely examples. Your solution seems to work well!

See below screenshots for some additional passed tests. Screenshots were without the translate.

<details><summary>Vertical Overflow</summary>

Screenshot 2024-08-15 at 1 34 52 PM
Screenshot 2024-08-15 at 1 35 06 PM


</details>

<details><summary>Tick Label</summary>

Screenshot 2024-08-15 at 1 39 08 PM
Screenshot 2024-08-15 at 1 37 44 PM



</details>

<details><summary>Flat Text</summary>


Screenshot 2024-08-15 at 1 40 40 PM
Screenshot 2024-08-15 at 1 42 10 PM



</details>

amattu2:
Commenting this so I don't forget. The SubmittedData tests keep causing Jest failures

https://github.com/CBIIT/crdc-datahub-ui/actions/runs/10409644376/job/28829712043?pr=441

src/components/NodeTotalChart/CustomTick.tsx
amattu2: Honestly...I'm not sure what I used it for originally or why I never checked without it. Thanks for finding this! a912b92

amattu2:
...And the file list dialog https://github.com/CBIIT/crdc-datahub-ui/actions/runs/10409644376/job/28829790535?pr=441

src/utils/statisticUtils.ts
amattu2: Definitely a good QoL suggestion. Added in 09a6ab7

src/components/NodeTotalChart/index.tsx
amattu2: Added in 435ddca

Alejandro-Vega (approved):
LGTM! Didn't retest, but code looks solid.

PR #447, opened 2024-08-27
CRDCDH-1444 QuickSight Dashboard Embedding 

amattu2:
### Overview

This PR integrates the AWS QuickSight SDK to build out a Operational Dashboard page. Additionally, it includes "minor" refactoring related to the navigation bar and router implementations.

> [!NOTE]
> This can be tested locally, but NGINX must be using the DEV tier, not DEV2.

(Ordered by relevance to the user story)

- Add Operation Dashboard page using AWS QuickSight SDK
   - Accessible to Admins, Data Curators, and Federal Leads
   - There is no Figma for the current page design as it's just temporary
- Migrate React LazyLoader and RequireAuth to independent components
  - Also added basic test coverage for these two components
- Refactor our navigation config to support role-based access 
  - Also did a minor code cleanup of navbar and added details to type definitions
  - Note for future – Our entire navigation/routing setup could use a lot more refactoring
- Delete unused "LogoTablet" component

### Related Ticket(s)

CRDCDH-1444 (FE Task)
CRDCDH-763 (Operations Page User Story)


src/content/operationDashboard/DashboardView.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const [embeddedDashboard, setEmbeddedDashboard] = useState<DashboardExperience>(null);
+  const dashboardElementRef = useRef<HTMLDivElement>(null);
+
+  const handleDashboardChange = (e: SelectChangeEvent) => {
+    setSearchParams({ type: e.target.value });
+    dashboardElementRef.current.innerHTML = "";
+    setEmbeddedDashboard(null);
+  };
+
+  const createEmbed = async () => {
```
Alejandro-Vega, line 89: This is mostly for development QOL, but could you add a guard to check if embeddedDashboard already is populated, if it is then return? The reason for this is when you make a change on this file and React hot reloads the page, then it will try to embed a duplicate iframe into the project which always fails. You can test this by loading the page, adding a console log, then viewing the page again without reloading. 

Screenshot 2024-08-29 at 12 18 09 PM

Alejandro-Vega (changes requested):
Everything looks solid! Works as expected and didn't find any major issues. I left two small requested changes, but they are very minor and one of them is unrelated to your implementation.  Great work! Also, thanks for your updates on the navigation!

The only thing I wish could be better is the donut charts look a bit blurry initially, but when you expand them they do look clean. I doubt there is a way of fixing this due to the way QuickSight works. Although, everything else looks great and everything is very customizable!

src/content/operationDashboard/DashboardView.tsx
```diff
@@ -0,0 +1,137 @@
+import { Box, FormControl, MenuItem, SelectChangeEvent, styled, Typography } from "@mui/material";
+import { isEqual } from "lodash";
+import { FC, memo, useEffect, useRef, useState } from "react";
+import { useSearchParams } from "react-router-dom";
+import {
+  createEmbeddingContext,
+  DashboardExperience,
+  FrameOptions,
+} from "amazon-quicksight-embedding-sdk";
+import StyledSelect from "../../components/StyledFormComponents/StyledSelect";
```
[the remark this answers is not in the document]
amattu2: Updated in 1f4f272. It doesn't look like it visually effects anywhere else.

src/content/operationDashboard/DashboardView.tsx
amattu2: I updated this to make the dashboard slightly more stable. This will have to be re-evaluated once we have additional dashboards, as I think we would need to rebuild the embeddedDashboard each time the selectbox changes. For now, this change should work fine though. e49ac8e

Alejandro-Vega (approved):
LGTM!

PR #452, opened 2024-08-29
CRDCDH-1498 Support CCDI Data Common

amattu2:
### Overview

This PR updates our supported Data Commons to introduce CCDI.

> [!NOTE]
> Model Navigator won't actually load the Data Model until we update our Data Models repo (CRDCDH-1465). It may be awhile before the model is ready, so this PR can be merged without that.

- Copy the CTDC Model Navigator configuration and update it to support CCDI

### Related Ticket(s)

CRDCDH-1498 (Task)
CRDCDH-1466 (User Story – Model Navigator)
CRDCDH-1476 (User Story – CCDI Data Submissions)


Alejandro-Vega (changes requested):
Config LGTM! Understood that this will not work until Data Models repo is updated. Although, it is currently throwing an uncaught runtime error when trying to load the page, instead of just showing the error message on the page. Kinda odd, shouldn't the ErrorBoundary be catching this error? 

Feel free to disregard as this will be fixed later with the data model update anyways.

amattu2:
> Config LGTM! Understood that this will not work until Data Models repo is updated. Although, it is currently throwing an uncaught runtime error when trying to load the page, instead of just showing the error message on the page. Kinda odd, shouldn't the ErrorBoundary be catching this error?
> 
> Feel free to disregard as this will be fixed later with the data model update anyways.

It only happens with the React dev server running, I think the intent is just to ensure the error isn't hidden during development? https://stackoverflow.com/a/52097626

Alejandro-Vega (approved):

PR #456, opened 2024-09-03
CRDCDH-1495 Support "Federal Monitor" Role

amattu2:
### Overview

This PR adds FE support for the new "Federal Monitor" role and includes some relevant refactoring of the profile management functionality.

- Move APITokenDialog to a separate components folder
- Define a separate type for the user roles (`UserRole`) instead of being defined in the User type
  - I did not replace every instance of `User["role"]` with `UserRole` to avoid a large changeset
- Migrate global type definitions for the Edit Profile/User inputs to the localized GraphQL mutation instead
- Major refactoring of the ProfileView page to improve QoL
  - Migrate form inputs to StyledFormComponents (slightly different spacing, but it's negligible)
  - ^^ Also a UX QA issue
  - Rewrite `getEditableFields` as a hook instead
  - Remove unnecessary `setFormValues` implementation (use `reset` directly)
  - Remove MUI error component and use a snackbar instead (the original implementation wasn't dictated by reqs.)
- Fix an issue with `formatIDP` throwing an exception for non-string input (rare issue caused by manual DB changes)

### Related Ticket(s)

CRDCDH-1495 (FE Task)
CRDCDH-1471 (User Story)


Alejandro-Vega (approved):
Everything LGTM and seems to work well. All the code looks solid, and good use of a hook! Great job!

PR #460, opened 2024-09-05
CRDCDH-1584 Navbar UX QA

amattu2:
### Overview

This PR addresses a small UX QA issue identified with the Navbar font weight.

- Update navbar item font-weight from 700 down to 600
- Remove unused font files that aren't referenced anywhere
- Migrate GenericAlert for auth errors to useSnackbar

### Related Ticket(s)

CRDCDH-1584 (Ticket)
CRDCDH-1515 (User Story)


src/components/Header/index.tsx
```diff
@@ -1,34 +1,29 @@
...
   const tabletAndMobile = useMediaQuery("(max-width: 1024px)");
 
-  const [showLoginError, setShowLoginError] = useState<boolean>(false);
-  const authContext = useAuthContext();
-
   useEffect(() => {
-    if (authContext.error !== undefined) {
-      setShowLoginError(true);
-      setTimeout(() => setShowLoginError(false), 10000);
+    if (typeof authError !== "string") {
```
Alejandro-Vega, line 15: Instead of returning here, should we instead display a generic error message, like an unknown authentication error has occurred or something like that? Since authError is still populating, it is still an error.

Alejandro-Vega (changes requested):
Everything LGTM! Left one small optional change.

src/components/Header/index.tsx
amattu2: Agreed. Thanks for catching this issue. 62a503e

Alejandro-Vega (approved):

PR #468, opened 2024-09-18
CRDCDH-1500 Create Approved Study List Page

Alejandro-Vega:
### Overview

Create a page for listing approved studies in the Admin portal. 

- Created Approved Study List route and page and modeled it after the Organization List View page
- Fixed retrieving approved studies to sort studies server-side instead of locally
- Added table displaying approved study data
- Disabled sorting for `Access Type` until the column requirements and BE are updated
- Provided filters to table
- URL search parameters are turned on for this table to save table state, including the filters
- Fixed filter bug for other tables where if you type into the input too fast then it will cause the page to crash. This was due to searchParams updating too quickly with the same values. Added a guard to prevent this.
- Added debounce to filters to avoid excess API calls

> [!NOTE]
> The `Access Type` column currently does not support sorting and is based on original requirements. This will be updated in the future PR.

### Related Ticket(s)

[CRDCDH-1500](https://tracker.nci.nih.gov/browse/CRDCDH-1500)


src/utils/studyUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @param {boolean} controlledAccess - Indicates whether controlled access is enabled.
+ * @param {boolean} openAccess - Indicates whether open access is enabled.
+ * @returns A formatted string listing the enabled access types separated by a comma and space.
+ *          - Returns "Controlled, Open" if both are enabled.
+ *          - Returns "Controlled" if only controlled access is enabled.
+ *          - Returns "Open" if only open access is enabled.
+ *          - Returns an empty string if neither is enabled.
+ */
+export const formatAccessTypes = (controlledAccess: boolean, openAccess: boolean) => {
```
amattu2, line 12: Based on the latest reqs, should this be "All" instead of "Controlled, Open"? Maybe that requirement was accidentally updated.

Also, if we're updating this function, it probably doesn't hurt to type the response too. Otherwise, disregard this.


```suggestion
export const formatAccessTypes = (controlledAccess: boolean, openAccess: boolean): string => {
```

amattu2 (approved):
There is one potential action item annotated below, otherwise everything in the FE looks good to me. I'm approving it just in case no update is needed.

BTW,

The API seems to handle filtering of the "Open" access type strangely. It is excluding studies with controlledAccess + openAccess data, but if you select "Controlled", it will show you studies with controlledAccess + openAccess data still. Definitely evaluate this and see if we need to reopen the BE task or if it's expected.

amattu2 (approved):
Reapproving based on clarification from KC. You can merge if there's no other changes needed.

PR #475, opened 2024-09-24
CRDCDH-1607 Data Submission List table improvements

Alejandro-Vega:
### Overview

Updated Data Submission List table to prevent text wrapping in the table rows.

- Reduced font size of all tables by 2px (16px => 14px)
- Reduced the start and end padding of all tables (40.44px => 24px)
- Set fixed-widths to specific columns in the Data Submission List table
- Forced text truncation for variable-width columns in the Data Submission List table. When truncated the text is hoverable and a tooltip will appear with full text
- Allowed wrapping in header names as this saves extra space
- If "Submission Name" columns is truncated, I left the original solid underline instead of changing it to a dashed underline to avoid confusion. Deleted/Archived submission do have the dashed underline as they are not clickable
- Moved "dbGaP ID" column to variable-length columns as it allows up to 50 characters in Submission Request form
- Updated dates to only show date portion, but will show full date and time on hover tooltip
- Tested with all letters and some emojis. Some unique characters with extra width could cause a scrollable table, but this is not likely to be used. (ex. "⸻" is a single character. LMK if you agree with ignoring these).

### Related Ticket(s)

[CRDCDH-1607](https://tracker.nci.nih.gov/browse/CRDCDH-1607) (Task)
[CRDCDH-1333](https://tracker.nci.nih.gov/browse/CRDCDH-1333) (User Story)

### DIFF
--- a/src/components/GenericTable/index.tsx
+++ b/src/components/GenericTable/index.tsx
@@ -48,10 +48,10 @@ const StyledTableContainer = styled(TableContainer)({
     background: "#E3EEF9",
   },
   "& .MuiTableCell-root:first-of-type": {
-    paddingLeft: "40.44px",
+    paddingLeft: "24px",
   },
   "& .MuiTableCell-root:last-of-type": {
-    paddingRight: "40.44px",
+    paddingRight: "24px",
   },
 });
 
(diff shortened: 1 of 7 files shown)


amattu2 (approved):
Everything LGTM! Nice job

PR #478, opened 2024-09-27
Hotfix: Fix Approved Studies queries and usage

Alejandro-Vega:
### Overview

The API was updated to support more functionality and params, but caused breaking changes to existing usage of this query. This PR is just to fix the existing usage.

- Updated `listApprovedStudies` query and `ApprovedStudy` type
- Updated return type for `ApprovedStudyOfMyOrganization` query
- Updated Profile page and Org view page to return all studies
- Fixed broken tests

### Related Ticket(s)

[CRDCDH-1500](https://tracker.nci.nih.gov/browse/CRDCDH-1500) (Related, but not the implementation)

### DIFF
--- a/src/content/users/ProfileView.tsx
+++ b/src/content/users/ProfileView.tsx
@@ -200,6 +200,12 @@ const ProfileView: FC<Props> = ({ _id, viewType }: Props) => {
   });
 
   const { data: approvedStudies } = useQuery<ListApprovedStudiesResp>(LIST_APPROVED_STUDIES, {
+    variables: {
+      // show all access types
+      controlledAccess: "All",
+      first: -1,
+      offset: 0,
+    },
     context: { clientName: "backend" },
     fetchPolicy: "cache-and-network",
     skip: fieldset.studies !== "UNLOCKED",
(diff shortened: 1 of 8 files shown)


Alejandro-Vega:
LMK if I missed any other breaking changes on the site.

src/content/users/ProfileView.tsx
```diff
@@ -208,7 +214,7 @@ const ProfileView: FC<Props> = ({ _id, viewType }: Props) => {
   // TODO: This is temporary until the API supports sorting natively
```
amattu2, line 214: I think this can be removed now since the API should support sorting by studyName?

src/types/Submissions.d.ts
```diff
@@ -381,3 +381,5 @@ type SubmitButtonResult = {
   tooltip?: string;
   _identifier?: string;
 };
+
+type AccessType = "All" | "Controlled" | "Open";
```
amattu2, line 385: I think this is technically a Application.d.ts (Questionnaire) type? Or maybe just a ApprovedStudy.d.ts type

amattu2 (approved):
I left two comments on things we may want to address in the future. Functionally this works as expected, so I'm going to merge.  

Thanks for putting this together.

PR #482, opened 2024-10-02
fix: Upgrade outdated action dependencies

amattu2:
### Overview

A QoL update to resolve GitHub Actions warnings related to relying on dependencies that use Node v16.


amattu2:
Skipping review. No code changes.

PR #485, opened 2024-10-08
CRDCDH-1611 Submission Request Cancer Types Autocomplete

amattu2:
### Overview

This PR migrates the existing Submission Request > Section C > Cancer Types dropdown to an autocomplete component, maintaining the multi-select functionality.

> [!NOTE]
> This migration is backwards-compatible with the previous implementation in 3.0.0.

- Build out a MUI Autocomplete wrapper which overwrites the native multi-select "tagging" functionality
  - We already have a autocomplete input for the form, but it's tightly constrained to the expected functionality of a "free-solo" (aka, any value is acceptable) autocomplete. I decided to create another one because I don't want to impact the existing implementation.
- Minor updates to Section C and how we handled the "Not Applicable" option selection

### Related Ticket(s)

CRDCDH-1611 (FE Task)
CRDCDH-1191 (User Story)

### DIFF
--- a/src/components/Questionnaire/CustomAutocomplete.tsx
+++ b/src/components/Questionnaire/CustomAutocomplete.tsx
@@ -90,12 +90,10 @@ const StyledAutocomplete = styled(Autocomplete)(({ readOnly }: { readOnly?: bool
       alignItems: "center",
       padding: "10.5px 30px 10.5px 12px !important",
       height: "44px",
-      ...(readOnly
-        ? {
-            backgroundColor: "#E5EEF4",
-            cursor: "not-allowed",
-          }
-        : {}),
+      ...(readOnly && {
+        backgroundColor: "#E5EEF4",
+        cursor: "not-allowed",
+      }),
     },
     "& .MuiInputBase-input": {
       padding: "0 !important",
@@ -286,9 +284,16 @@ const CustomAutocomplete = ({
         />
         <StyledHelperText>{!readOnly && error ? helperText : " "}</StyledHelperText>
       </StyledFormControl>
-      <ProxySelect name={name} aria-labelledby={`${id}-label`} multiple hidden>
+      <ProxySelect
+        name={name}
+        value={val}
+        aria-labelledby={`${id}-label`}
+        onChange={() => {}}
+        multiple
+        hidden
+      >
         {val.map((v) => (
-          <option key={v} value={v} aria-label={v} selected />
+          <option key={v} value={v} aria-label={v} />
         ))}
       </ProxySelect>
     </Grid>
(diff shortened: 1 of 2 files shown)


src/components/Questionnaire/CustomAutocomplete.tsx
```diff
@@ -0,0 +1,298 @@
...
+              }}
+            />
+          )}
+          {...rest}
+        />
+        <StyledHelperText>{!readOnly && error ? helperText : " "}</StyledHelperText>
+      </StyledFormControl>
+      <ProxySelect name={name} aria-labelledby={`${id}-label`} multiple hidden>
+        {val.map((v) => (
+          <option key={v} value={v} aria-label={v} selected />
```
Alejandro-Vega, line 291: This "selected" prop is causing an error, see below.

image


I quickly grabbed this from the proxy select for the SelectInput component also used in the questionnaire, but it seems to work the same without the error. Feel free to change it how you see fit:

```javascript
<ProxySelect
  value={val}
  onChange={() => {}}
  name={name}
  aria-labelledby={`${id}-label`}
  multiple
  hidden
  >
    {val.map((v) => (
      <option key={v} value={v} aria-label={v} />
    ))}
</ProxySelect>
```

src/components/Questionnaire/CustomAutocomplete.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+const StyledAutocomplete = styled(Autocomplete)(({ readOnly }: { readOnly?: boolean }) => ({
+  "& .MuiInputBase-root": {
+    "&.MuiAutocomplete-inputRoot.MuiInputBase-root": {
+      display: "flex",
+      alignItems: "center",
+      padding: "10.5px 30px 10.5px 12px !important",
+      height: "44px",
+      ...(readOnly
+        ? {
```
Alejandro-Vega, line 94: Ternary is unnecessary, you can use the && operator for the same result:

```javascript
...(readOnly && {
        backgroundColor: "#E5EEF4",
        cursor: "not-allowed",
}),
```

Alejandro-Vega (changes requested):
Everything works well. Just left some small suggested changes. Great work!

src/components/Questionnaire/CustomAutocomplete.tsx
amattu2: Thanks. Updated in 01753b7.

src/components/Questionnaire/CustomAutocomplete.tsx
amattu2: Thanks for catching this, I always forget to check the console. Updated in 01753b7

Alejandro-Vega (approved):
LGTM!

PR #487, opened 2024-10-10
CRDCDH-1686 Base implementation for Emoji Validation

amattu2:
### Overview

This PR "fixes" a few places where emojis were implicitly allowed, it also makes some related changes to our TS project config.

- Sync ESLint and TypeScript ECMA version to ES2015 (ES6)
  - I bumped the ES version to ES6 since that was required to use a unicode RegEx. The browser support for the FE drops slightly, but it's most likely worth the trade-off
- (Unrelated) remove explicit type injection for TS, and use defaults
- Create a emoji validation utility function that integrates with React Hook Form (also kinda works with the Questionnaire)
- Prevent emojis in:
  - Create Data Submission > dbGaPID 
  - Create Data Submission > Submission Name
  - Submission Request > Section B > Study Name (This was only per Sofia request, but all other fields are completely out of scope of this bug)
- Update/create related test coverage

### Related Ticket(s)

CRDCDH-1686

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -21,7 +21,7 @@ const config = {
     ecmaFeatures: {
       jsx: true,
     },
-    ecmaVersion: 11,
+    ecmaVersion: 2015,
     project: "./tsconfig.json",
     sourceType: "module",
   },
--- a/src/components/DataSubmissions/CreateDataSubmissionDialog.test.tsx
+++ b/src/components/DataSubmissions/CreateDataSubmissionDialog.test.tsx
@@ -896,4 +896,124 @@ describe("Basic Functionality", () => {
       expect(createButton).not.toBeInTheDocument();
     });
   });
+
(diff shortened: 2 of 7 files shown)


Alejandro-Vega (approved):
@amattu2  Everything LGTM, but I do notice the error message is different in the questionnaire than in the dialog. The ticket doesn't really describe the error message so I will leave it at your discretion if you would like to address this. Feel free to merge if you will leave it as-is.

amattu2:
> @amattu2 Everything LGTM, but I do notice the error message is different in the questionnaire than in the dialog. The ticket doesn't really describe the error message so I will leave it at your discretion if you would like to address this. Feel free to merge if you will leave it as-is.

Another unfortunate byproduct of the fact that we're using basically custom coded form management on the Submission Request form. To minimize changes to the submission request page/components, I'll leave this alone for now. Ideally it would use the same error message across the board for emojis.

PR #488, opened 2024-10-10
CRDCDH-1675 Fix Other Clinical Data Types delimiter

amattu2:
### Overview

Minor PR to fix a 3.0.0 regression for the Other Clinical Data Types delimiter. This was introduced when merging 2.1.0 into 3.0.0

- Remove the explicit `split` call on the Other Clinical Data Types field in the Review Section. Calling `split` on the field pre-maturely turned it into an array, causing the `|` delimiter to be unused

### Related Ticket(s)

CRDCDH-1675


Alejandro-Vega (approved):
LGTM, pipe delimiter seems to be working as expected.

PR #491, opened 2024-10-10
Hotfix: CRDCDH-1609 Data Submission Filters - Fix form from updating with same value 

Alejandro-Vega:
### Overview

Fix the debouncing not working when the organization dropdown is read only. This is due to the form updating with the same organization value every time there's a change.

- Added check to make sure the form doesn't update with the same value for organization

### Related Ticket(s)

[CRDCDH-1609](https://tracker.nci.nih.gov/browse/CRDCDH-1609) (Task)
[CRDCDH-1412](https://tracker.nci.nih.gov/browse/CRDCDH-1412) (User Story)


amattu2 (approved):
LGTM. Didn't test

PR #496, opened 2024-10-15
CRDCDH-1497

ty-esi:
### Overview

Data Model Navigator improvements to support CDEs. 

- refactored `updateEnums` to add CDE info (fullName, version, etc) to the dictionary object for each property.
- updated menu names under available downloads.

### Related Ticket(s)

https://tracker.nci.nih.gov/browse/CRDCDH-1497


amattu2 (approved):
Both FE and DMN changes LGTM. I updated the package in commit dbf24af and also added a missing attribute so the links build.

Nice work!

PR #499, opened 2024-10-17
CRDCDH-1743 Implement "Request Access" form

amattu2:
### Overview

This PR introduces a new feature for external users (User, Submitter, Org Owner) to request a role and/or organization change via their profile.

It also incorporates related DX QoL changes. See details below.

> [!NOTE]
> I added a "StyledDialogComponents" folder, but did not migrate any existing dialogs to these prestyled components. If we get related work (bugs/tasks) for the dialogs, we can probably migrate them at that point.
>
> Most of the dialogs have the same general design, with the exception of width and content-specific customizations. If it doesn't make sense to have a centralized set of components, I can remove this folder.
>
> Related note – I left the explicit font specification on the dialog components, if any dialog can appear on the Model Navigator page (eg. CLI Tool dialog), Model Navigator will try to override our default Nunito font.

> [!Warning]
> This PR has several tests that refuse to pass in CI consistently. I plan to revisit this in #511 as I'm unsure what the root issue is ATM.

- Add a "Request Access" form for certain external user roles to request a specific role and organization assignment
- Deduplicate the Autocomplete input field styling and create a component for it in the StyledFormComponents folder
- Add the standard dialog components to a generic folder (similar to StyledFormComponents)
- Create a new query specifically to fetch organization names, which is heavily cached once fetched

### Related Ticket(s)

CRDCDH-1743 (Task)


amattu2:
NOTE: The additional input field has a false-positive 508 issue. This is the same issue that is described here #510. However, because this field does need to be multi-line, I can't find a solution for it.

Screenshot 2024-10-29 at 3 17 21 PM

src/components/HistoryDialog/index.test.tsx
```diff
@@ -21,13 +20,14 @@ const BaseProps: React.ComponentProps<typeof HistoryDialog> = {
   onClose: () => {},
 };
 
-describe("Accessibility", () => {
-  it("should have no violations", async () => {
-    const { container } = render(<HistoryDialog {...BaseProps} />);
+// TODO: Fix this failing in CI
```
Alejandro-Vega, line 23: I tried to run this test on my machine several times, and it seems to be passing.. Maybe I just haven't had it run enough.

Alejandro-Vega (dismissed):
Everything LGTM! I like the centralized styled components for the dialog. Thanks for adding that. I don't have any actual requested changes, just some comments.

src/components/AccessRequest/FormDialog.tsx
```diff
@@ -0,0 +1,288 @@
...
+ * @param {Props} props
+ * @returns {React.FC<Props>}
+ */
+const FormDialog: FC<Props> = ({ onClose, ...rest }) => {
+  const { user } = useAuthContext();
+  const { enqueueSnackbar } = useSnackbar();
+
+  const { handleSubmit, register, control, formState } = useForm<InputForm>({
+    defaultValues: {
+      role: RoleOptions.includes(user.role) ? user.role : "Submitter",
```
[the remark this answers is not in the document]
amattu2: Completely agree, I didn't even think about this.. 

There's also no rate limit, so a user can just keep submitting it over and over..

src/components/HistoryDialog/index.test.tsx
amattu2: This is why it's so hard to debug it, it hasn't failed a single time locally for me. The HistoryDialog is also doing the same thing. All of them keep failing for an act warning.

amattu2:
Skipping re-review, a merge commit dismissed the previous approval.

PR #501, opened 2024-10-18
CRDCDH-1740 Redesign History Dialog component

amattu2:
### Overview

This PR introduces a rewrite of the History Dialog's underlying component, which incorporates a design update and an additional field indicating which user performed the history transition.

> [!NOTE]
> This design change impacts the Submission Request too, but the "User" column won't appear there (intentionally) since we don't get `userName` from the getApplication API.

- Rewrite the History Dialog to be slightly more efficient (memoization and optimizing usage of callbacks, etc)
- Rebuild the dialog UX without the MUI timeline (we essentially weren't using anything from it)
- Add support for custom styling on the TruncatedText component
- Minor updates to unit tests that were impacted by changes
- ~~Attempt to resolve transition-based ACT warnings by disabling MUI transitions~~ [Did not work](https://github.com/CBIIT/crdc-datahub-ui/actions/runs/11483531612/attempts/1)

### Related Ticket(s)

CRDCDH-1740 (FE Task)
CRDCDH-1741 (Design)
CRDCDH-1697 (User Story)

### DIFF
--- a/src/components/HistoryDialog/index.tsx
+++ b/src/components/HistoryDialog/index.tsx
@@ -76,13 +76,13 @@ const StyledCloseButton = styled(Button)({
   },
 });
 
-const StyledGridHeader = styled(Grid)({
+const StyledHeaderRow = styled(Grid)({
   borderBottom: "0.5px solid #375F9A",
   paddingBottom: "8px",
   marginBottom: "4px",
 });
 
-const StyledHistoryHeader = styled(Typography)<React.CSSProperties>((styles) => ({
+const StyledHeaderItem = styled(Typography)<React.CSSProperties>((styles) => ({
   fontFamily: "Public Sans",
   fontWeight: "300",
   fontSize: "8px",
@@ -93,7 +93,11 @@ const StyledHistoryHeader = styled(Typography)<React.CSSProperties>((styles) => 
   ...styles,
 }));
 
-const StyledGridEventItem = styled(Grid)({
+const HeaderGap = styled("div")({
+  height: "40px",
+});
+
+const StyledEventRow = styled(Grid)({
   padding: "17px 0",
   borderBottom: "0.5px solid #375F9A",
   alignItems: "center",
@@ -107,7 +111,7 @@ const BaseItemTypographyStyles: React.CSSProperties = {
   userSelect: "none",
 };
 
-const StyledHistoryItem = styled(Typography)<React.CSSProperties>(
+const StyledEventItem = styled(Typography)<React.CSSProperties>(
   ({ textAlign = "center", color = "inherit" }) => ({
     ...BaseItemTypographyStyles,
     textAlign,
@@ -189,11 +193,38 @@ type EventItem = {
 export type IconType<T extends string> = Record<T, string>;
 
 type Props<T extends string> = {
+  /**
+   * The prefix to show before the dialog title
+   */
   preTitle: string;
+  /**
+   * The title of the dialog
+   */
   title: string;
+  /**
+   * The history items to display
+   *
+   * @see {@link HistoryBase}
+   */
   history: HistoryBase<T>[];
+  /**
+   * A map of icons to display for each status
+   *
+   * @note If iconMap[status] is undefined, no icon will be displayed
+   */
   iconMap: IconType<T>;
+  /**
+   * Boolean indicator of whether to render the headers of each column
+   */
+  showHeaders?: boolean;
+  /**
+   * A function to determine the text color of the status
+   */
   getTextColor: (status: T) => CSSProperties["color"];
+  /**
+   * A function to call when the dialog is requested to close by an event,
+   * e.g. Close button click, backdrop click, or escape key press
+   */
   onClose: () => void;
 } & DialogProps;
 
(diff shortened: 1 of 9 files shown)


src/components/HistoryDialog/index.tsx
```diff
@@ -51,149 +49,236 @@ const StyledTitle = styled("p")({
...
+  content: '""',
+  position: "absolute",
+  left: "5px",
+  top: "0",
+  width: "6px",
+  height: "27px",
+  background: "white",
+});
+
+const HorizontalLine = styled("div")({
```
Alejandro-Vega, line 160: The horizontal line is over the timeline track.

Screenshot 2024-10-23 at 10 58 59 AM

src/components/HistoryDialog/index.tsx
```diff
@@ -208,53 +293,105 @@ const HistoryDialog = <T extends string>({
...
-                      />
-                    </StyledAvatar>
-                  )}
-                </Stack>
-              </StyledTimelineContent>
-            </StyledTimelineItem>
-          ))}
-        </Timeline>
+              {icon && (
+                <StyledIcon>
```
Alejandro-Vega, line 377: The icon seems a little outside the bounds compared to the design for Data Submissions. The Submission Request dialog looks fine.

Current:
Screenshot 2024-10-23 at 10 53 27 AM

UI Design:
Screenshot 2024-10-23 at 10 53 39 AM

src/components/HistoryDialog/index.tsx
```diff
@@ -51,149 +49,236 @@ const StyledTitle = styled("p")({
 const StyledDialogContent = styled(DialogContent)({
```
Alejandro-Vega, line 49: Vertical spacing between title and content, as well as content and Close button seems smaller than then UI design. As well as under the Close button.

Current:
Screenshot 2024-10-23 at 11 35 48 AM

UI Design:
Screenshot 2024-10-23 at 11 35 42 AM

Alejandro-Vega (changes requested):
Super minor styling changes suggested. Feel free to disregard any if you feel otherwise.

src/components/HistoryDialog/index.tsx
amattu2: I'm not seeing this on Chrome v130, but I applied a z-index to the dot container in commit 5a27dad. I'm going to leave this unresolved, LMK if the issue is still present and I'll try something else. 

HR Overlap

src/components/HistoryDialog/index.tsx
amattu2: Adjusted accordingly in the commit 5a27dad. I think the spacing is still slightly off, but it's an improvement. Please feel free to unresolve this if you want it adjusted more.

src/components/HistoryDialog/index.tsx
Alejandro-Vega: The issue still remains, but is inconsistently showing. There seems to be a small gap between the TopConnector and BottomConnector which makes it so the border is sometimes showing. (Unsure why only sometimes) But if I zoom in 200% on Chrome and increase the border size to 5px, it becomes more noticeable. 

Screenshot 2024-10-24 at 12 49 13 PM

I'm sure it is possible to accomplish a fix using box-sizing or negative margins, but I tried them and they didn't seem to work without a lot of tweaking. They would obviously be a better solution, so if you're able to figure it out, then even better. Feel free to disregard my solution below, if you find a better one! It was just my initial thoughts.

A working, but not ideal, solution I found was just setting a CSS variable for the border-width in the `StyledDialogContent` where it can be shared, then using that to calculate the connector heights:

```javascript
const StyledDialogContent = styled(DialogContent, { shouldForwardProp: (p) => p !== "headerRow" })<{
  headerRow: boolean;
}>(({ headerRow }) => ({
  "--border-bottom-width": "1px",  // I recommend using 1px instead of 0.5px, on a screen they are equivalent
  
  ...
  
}));

```

Then we can use this wherever we define our borders:

```javascript
const StyledHeaderRow = styled(Grid)({
  borderBottom: "var(--border-bottom-width) solid #375F9A",
  
  ...
  
});

const StyledEventRow = styled(Grid)({
  borderBottom: "var(--border-bottom-width) solid #375F9A",
  
  ...
  
});
```


Finally, just extend the connect height by half the border for both TopConnector and BottomConnector
```javascript
const TopConnector = styled("div")({
  height: "calc(27px + var(--border-bottom-width) / 2)",
  
  ...
  
});

const BottomConnector = styled("div")({
  height: "calc(27px + var(--border-bottom-width) / 2)",
  
  ...
  
});
```

Result:
Screenshot 2024-10-24 at 1 26 29 PM

Screenshot 2024-10-24 at 1 24 36 PM

amattu2: Thanks for the advice on using `5px`, I can see it clearly with any zoom at 5px. For some reason, even at 250% zoom on `0.5px`, the gap wasn't visible. Regardless, I like your solution and implemented it in commit 8eee7e8.

Interestingly, `0.5px` renders differently for me than `1px`, so I left it at `0.5px`. I'm assuming the reason it's rendering differently for me vs you is also the reason you can see the gap and I can't.

<details><summary>0.5px vs 1.0px</summary>
<p>

(I don't know whether this will actually show up well in a screenshot 😃)

`0.5px`

Half-Px-Border

`1px`

1-Px-Border

</p>
</details>

Alejandro-Vega (approved):
LGTM!

PR #503, opened 2024-10-23
CRDCDH-1826 Add a fake data model

amattu2:
### Overview

This PR introduces a fake data model "Hidden Model" so that we can test the hiding functionality on lower tiers without using a real data model. After a DevOps change, this model will be hidden permanently.

> [!Note]
> This model does not have any actual model files defined ATM, so it will not load in model navigator.

- Add a fake data model with bare minimum configuration
- Change the NavBar config to not append "Model" after the DC name if it includes "Model" already (e.g. "Hidden Model Model")

### Related Ticket(s)

CRDCDH-1826 (FE Task)

### DIFF
--- a/src/config/DataCommons.ts
+++ b/src/config/DataCommons.ts
@@ -332,6 +332,42 @@ const DataCommons: DataCommon[] = [
     },
   },
   {
+    name: "Hidden Model",
+    assets: null,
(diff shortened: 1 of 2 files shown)


Alejandro-Vega (approved):
LGTM!

PR #515, opened 2024-10-30
CRDCDH-1882 QuickSight – Role-Based Content Filtering

amattu2:
### Overview

This PR updates the QuickSight implementation to provide contextual information to the dashboard. In particular, it provides the assigned list of Studies or Data Commons assigned to the user, which QuickSight then uses to filter the Data Submission Dashboard content. 

Related to this, I also updated the error handling of the Data Submission page to be consistent with the Submission Request Page, Organization Page, Profile Page, etc by redirecting back to the Data Submission List page.

> [!NOTE]
> The Data Submission page still has 1 inconsistency with most of the other features – There is no full screen loader while fetching the submission data initially. I didn't address this as it may have significant side-effects, but it's something we might want to address eventually. 

- Revamp the DashboardView to include additional Data Commons or Study parameters
- Log edge cases when the Federal Monitor or Data Curator don't have studies or data commons assigned and fallback to empty content
  - NOTE: Technically it's acceptable for a Fed Monitor to not have any studies, but QuickSight interprets [] as "show everything, which we definitely don't want.
- Update the Data Submission page to handle errors consistent with the rest of the app (redirect to list page)
  - NOTE: I hard-coded the error message because the APIs return very obscure error messages that have no value to the user (IMO)
- Update the submission action error handler to show a snackbar instead of a error at the top of the data submission summary
- Update the Data Submission Ctx to log any errors using the Logger util (for debugging)

### Related Ticket(s)

CRDCDH-1882 (FE Task)
CRDCDH-1881 (User Story 2)
CRDCDH-1880 (User Story 1)
CRDCDH-1848 (Bug 1)
CRDCDH-1849 (Bug 2)

### DIFF
--- a/src/components/Contexts/SubmissionContext.tsx
+++ b/src/components/Contexts/SubmissionContext.tsx
@@ -2,7 +2,7 @@ import React, { FC, createContext, useCallback, useContext, useMemo, useState } 
 import { ApolloError, ApolloQueryResult, useQuery } from "@apollo/client";
 import { cloneDeep, isEqual } from "lodash";
 import { GetSubmissionResp, GET_SUBMISSION, GetSubmissionInput } from "../../graphql";
-import { compareNodeStats } from "../../utils";
+import { compareNodeStats, Logger } from "../../utils";
 
 export type SubmissionCtxState = {
   /**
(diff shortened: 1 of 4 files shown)


Alejandro-Vega (approved):
LGTM! Thanks for adding all of the improved error handling. The redirect with the error message also seems to work well.

PR #516, opened 2024-10-31
CRDCDH-1738 Data Submission List - Unlock Organization dropdown

Alejandro-Vega:
### Overview

Update the Organization select dropdown from the Data Submission List table filters so that any user can select from relevant Organizations.

- Updated listSubmissions API to retrieve Organizations
- Updated Data Submission List filters to not use OrganizationListContext, and replace with orgs from listSubmissions API
- Updated Org dropdown to reset to "All" when the selected org is no longer an option
- After a new submission is created, update the orgs, submitter names, and dataCommons lists
- Updated tests to match new implementation
- Fixed lineHeight issue in Data Submission summary section. The collaborators property was taller than all the other properties
- Added a bandage fix for an existing bug. Unsure of cause, but please test to ensure it is fixed for you as well. Added comment with more details below
 
### Related Ticket(s)

[CRDCDH-1639](https://tracker.nci.nih.gov/browse/CRDCDH-1639) (Task)
[CRDCDH-1639](https://tracker.nci.nih.gov/browse/CRDCDH-1639) (User Story)

### DIFF
--- a/src/graphql/listSubmissions.ts
+++ b/src/graphql/listSubmissions.ts
@@ -73,7 +73,7 @@ export type Response = {
   listSubmissions: {
     total: number;
     submissions: Omit<Submission, "submitterID" | "bucketName" | "rootPath" | "history">[];
-    organizations: Organization[];
+    organizations: Pick<Organization, "_id" | "name">[];
     submitterNames: string[];
     dataCommons: string[];
   };
--- a/src/graphql/listSubmissions.ts
+++ b/src/graphql/listSubmissions.ts
@@ -46,6 +46,10 @@ export const query = gql`
         updatedAt
         intention
       }
+      organizations {
+        _id
+        name
+      }
       submitterNames
       dataCommons
     }
@@ -69,6 +73,7 @@ export type Response = {
   listSubmissions: {
     total: number;
     submissions: Omit<Submission, "submitterID" | "bucketName" | "rootPath" | "history">[];
+    organizations: Organization[];
     submitterNames: string[];
     dataCommons: string[];
   };
--- a/src/components/DataSubmissions/DataSubmissionListFilters.test.tsx
+++ b/src/components/DataSubmissions/DataSubmissionListFilters.test.tsx
@@ -10,7 +10,6 @@ import {
   ContextState as AuthContextState,
   Status as AuthContextStatus,
 } from "../Contexts/AuthContext";
-import { useOrganizationListContext } from "../Contexts/OrganizationListContext";
 import { Column } from "../GenericTable";
 import { ListSubmissionsResp } from "../../graphql";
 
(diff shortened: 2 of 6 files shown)


Alejandro-Vega:
@amattu2  The existing bug was where if you populate all dropdown fields to something other than "All", then switch the page, then switch one of the three dropdowns Org, Data Commons, or Submitter Names, it would switch to the new value but then switch back to the original value like so: "Alex" => "All" => "Alex". You should be able to see this on DEV. I believe this is a race-condition with the setting of the filters and resetting the table page back to the first. 

Testing Steps:
- Populate all dropdown fields to something other than "All"
- Change to a page other than the first one
- Modify a dropdown filter (Org, Data Commons, or Submitter Names) to something else
- It should change the page back to 1 and the dropdown filter should change to the one you selected

Somehow adding a debounce with a 0 timer fixes the issue? Unsure how, but that's all I could come up with for now. If you have a better solution, please share. I've been debugging this for a while, but all my other solutions break something else unfortunately.

src/graphql/listSubmissions.ts
```diff
@@ -69,6 +73,7 @@ export type Response = {
   listSubmissions: {
     total: number;
     submissions: Omit<Submission, "submitterID" | "bucketName" | "rootPath" | "history">[];
+    organizations: Organization[];
```
amattu2, line 76: Technically this is `Pick<Organization, "_id" | "name">`, but this isn't a big deal ATM.

amattu2:
> @amattu2 The existing bug was where if you populate all dropdown fields to something other than "All", then switch the page, then switch one of the three dropdowns Org, Data Commons, or Submitter Names, it would switch to the new value but then switch back to the original value like so: "Alex" => "All" => "Alex". You should be able to see this on DEV. I believe this is a race-condition with the setting of the filters and resetting the table page back to the first.
> 
> Testing Steps:
> 
> * Populate all dropdown fields to something other than "All"
> * Change to a page other than the first one
> * Modify a dropdown filter (Org, Data Commons, or Submitter Names) to something else
> * It should change the page back to 1 and the dropdown filter should change to the one you selected
> 
> Somehow adding a debounce with a 0 timer fixes the issue? Unsure how, but that's all I could come up with for now. If you have a better solution, please share. I've been debugging this for a while, but all my other solutions break something else unfortunately.

I can replicate this somewhat consistently on DEV following your steps. Some observations:

- The URL correctly updates by removing the filter, even though the filter itself doesn't reflect that
- The Data Activity tab doesn't have this issue (also uses debouncing, so wanted to check)
- `handleFormChange` is called 3 times under this scenario
  1. Correct values
  2. Old values
  3. Old values
 
I spent some time looking into it, but I can't think of what the root cause would be ATM. With your solution, it seems to be smooth and stable enough to just leave as-is for now.

amattu2 (dismissed):
I left one annotation for a type issue, but it's not that significant. Approving as-is, if you decide to update it, feel free to merge without another approval.

src/graphql/listSubmissions.ts
Alejandro-Vega: Agreed, updated in [99a458b](https://github.com/CBIIT/crdc-datahub-ui/pull/516/commits/99a458b2d8e1a512ce26fe037fc351052e1e76e4).

PR #521, opened 2024-11-01
Synchronize CDE parsing logic

amattu2:
### Overview

This PR updates the CDE integration to sync the parsing logic. Specifically:

- Empty permissible values array – convert the property to a string
- API does not return anything for a specific CDE (it's not in the API response) – show an error for that CDE
- Search for the `Term` with caDSR origin, instead of just the first one (DH requirement change)
- Always show the CDE Origin from the MDF, even though it will always (in 3.1.0) be "caDSR" or akin

The first two logic updates are highlighted below, everything else in the table was already covered.

<details><summary>Details</summary>
<p>

Screenshot 2024-11-01 at 4 05 20 PM

</p>
</details> 

- Always populate CDE info (even if there's an issue with the PermissibleValues)
- Use the exact "CDE Origin" provided in the MDF for the caDSR CDE Term
- Update logic to match the matrix above
- Log an error when the CDE cannot be mapped

### Related Ticket(s)

N/A – Related to CRDCDH-1451

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8897,7 +8897,7 @@
     },
     "node_modules/data-model-navigator": {
       "version": "1.1.34",
-      "resolved": "git+ssh://git@github.com/CBIIT/Data-Model-Navigator.git#4b5c21a892c84178088541f9a31717d74f048087",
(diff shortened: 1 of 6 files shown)


amattu2:
Updated to reflect discussion with Ming. Additional changes:

- Use whatever the exact CDE origin is from the MDF model

Alejandro-Vega (approved):
LGTM!

PR #525, opened 2024-11-07
Restyle metric field per UX

amattu2:
### Overview

Minor UX updates to the Operation Dashboard metric dropdown.

- Restyle the Operation Dashboard page to remove negative margins for positioning
- Adjust the dropdown and label styling per UX
- Fix the QuickSight frame not rendering the rounded border

### Related Ticket(s)

CRDCDH-1428 (Design task)

### DIFF
--- a/src/content/operationDashboard/DashboardView.tsx
+++ b/src/content/operationDashboard/DashboardView.tsx
@@ -9,6 +9,7 @@ import {
   FrameOptions,
 } from "amazon-quicksight-embedding-sdk";
 import StyledSelect from "../../components/StyledFormComponents/StyledSelect";
+import StyledLabel from "../../components/StyledFormComponents/StyledLabel";
 import SuspenseLoader from "../../components/SuspenseLoader";
 import bannerSvg from "../../assets/banner/submission_banner.png";
 import { useAuthContext } from "../../components/Contexts/AuthContext";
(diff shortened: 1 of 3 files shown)


src/content/operationDashboard/DashboardView.tsx
```diff
@@ -9,6 +9,7 @@ import {
   FrameOptions,
 } from "amazon-quicksight-embedding-sdk";
 import StyledSelect from "../../components/StyledFormComponents/StyledSelect";
```
Alejandro-Vega, line 11: There are some weird overflowing corners after you click the dropdown then click away to close the dropdown and the input is still in focus.

Screenshot 2024-11-08 at 9 45 16 AM

I think I found the cause, this seems to fix it. Could you add something like this directly to the StyledSelect component?

```javascript
"&.Mui-focused .MuiOutlinedInput-input:focus": {
    borderRadius: "8px",
},
```

src/content/operationDashboard/DashboardView.tsx
```diff
@@ -20,45 +21,42 @@ export type DashboardViewProps = {
   loading: boolean;
 };
 
-const StyledViewHeader = styled(Box)({
+const StyledPageContainer = styled(Box)({
   background: `url(${bannerSvg})`,
```
Alejandro-Vega, line 25: The background in the Figma seems to be different. Are we going with the one you are currently using?

Alejandro-Vega (changes requested):
Mostly LGTM! One minor change and one question.

src/content/operationDashboard/DashboardView.tsx
amattu2: That's some really odd behavior. Thanks for catching it, fixed in b8cd159.

src/content/operationDashboard/DashboardView.tsx
amattu2: TBD, but it doesn't seem like it. If I get an answer, I'll update it in the future.

Alejandro-Vega (approved):
Didn't re-test but code LGTM!

PR #526, opened 2024-11-08
CRDCDH-1976 Fix Node labels overlapping

amattu2:
### Overview

PR to apply truncation to the Individual Pie Chart labels after 14 characters. I tried a few different combinations and 14 characters seems reasonable, but if there's a number that visually looks better, LMK.

<details><summary>New vs. Old</summary>
<p>

Screenshot 2024-11-08 at 10 43 52 AM

Screenshot 2024-11-08 at 10 43 13 AM

</p>
</details> 

- Apply standard truncation when a node label exceeds 14 characters
- Update how we handle label reformatting to try and closer-respect what the MDF dictates
- Add test coverage for label handling

### Related Ticket(s)

CRDCDH-1976

### DIFF
--- a/src/components/NodeChart/index.tsx
+++ b/src/components/NodeChart/index.tsx
@@ -3,6 +3,8 @@ import { Box, Typography, styled } from "@mui/material";
 import { PieChart, Pie, Label, Cell } from "recharts";
 import { isEqual } from "lodash";
 import PieChartCenter from "./PieChartCenter";
+import TruncatedText from "../TruncatedText";
+import { capitalizeFirstLetter, titleCase } from "../../utils";
 
 type Props = {
   /**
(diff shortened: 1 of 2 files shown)


src/components/NodeChart/index.tsx
```diff
@@ -53,16 +51,38 @@ const NodeChart: FC<Props> = ({ label, centerCount, data }: Props) => {
...
     () => (dataset.length === 0 && hoveredSlice === null) || hoveredSlice?.value === 0,
     [dataset, hoveredSlice]
   );
 
+  const reformattedLabel = useMemo<string>(() => {
+    const replacedLabel = label?.replace(/_/g, " ") || "";
+
+    // If the label has no spaces, capitalize the first letter to avoid
+    // titleCase from performing a full title case conversion
+    if (replacedLabel?.indexOf(" ") === -1) {
```
amattu2, line 65: Just some additional context:

`titleCase` forces all characters to lowercase first before applying uppercasing to the words. 

This works fine when a node name has multiple words (Example: `Principal Investigator` or `Non Targeted Therapy`) but it's pretty bad for the new nodes that CDS just created (Example: `NonDICOMCTimages` which would become `Nondicomctimages`).

My solution ATM is just to short-circuit the reformatting when there are no spaces. We never really had requirements that dictated how we format node labels, so it's somewhat inconsistent across the board.

Alejandro-Vega (approved):
LGTM!

PR #534, opened 2024-11-14
CRDCDH-2025 Updating an existing collaborator can cause field clearing

amattu2:
### Overview

This PR addresses an issue where:

1. A completely valid collaborator is added to a submission
2. The collaborator becomes invalid via some action, such as
  - Their role changes
  - Their org changes
  - A study is removed from their org
3. When the submitter goes to the manage collaborators table and changes their permission
4. The Name and Organization fields go blank until saved and reloaded

- If the user does not exist in the potential collaborators, backfill the information from the existing collaborator assignment
- Add an error log for when this occurs – This can be removed if desired
- Update test coverage for this scenario

### Related Ticket(s)

CRDCDH-2025

### DIFF
--- a/src/components/Contexts/CollaboratorsContext.test.tsx
+++ b/src/components/Contexts/CollaboratorsContext.test.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { fireEvent, render, waitFor } from "@testing-library/react";
+import { act, fireEvent, render, renderHook, waitFor } from "@testing-library/react";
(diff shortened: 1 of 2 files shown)


amattu2:
@Alejandro-Vega I'm merging this now due to Dev Complete (tomorrow). This is somewhat of an ugly hotfix, so when you return please take a look and see if we want to improve it or address it some other way (depending on urgency, either in 3.1.0 or 3.2.0).

PR #27, opened 2023-07-03
CRDCDH-36 Implement Section C

Alejandro-Vega:
Overview
This PR is the implementation of Form Section C for the ticket [CRDCDH-36](https://tracker.nci.nih.gov/browse/CRDCDH-36)

PR #536, opened 2024-11-18
CRDCDH-1861 Submission Request Content Adjustments

amattu2:
### Overview

This PR reorganizes some of the Submission Request content and includes minor QoL updates for the end users.

> [!NOTE]
> This change, specifically the moving of the dbGaPID field, should be fully backwards compatible. I tested working on the same form "simultaneously" using DEV2 and locally, but it's possible I missed something when testing.

Requirement based changes:

- Move dbGaPID field to Section C
  - Since dbGaPID is serialized inside the Study details, there's some migration code to ensure study data is properly stored.
  - Another option would be to pull dbGaPID out of study – Open to opinions on this
- Adjust max length of Funding Agency -> Grant field (Section B)
- Reorganize the last 3 inputs to make the Additional Comments box the last field in the form (Section D)
- Apply reorganization on the Review page (not explicitly called for, but inferred)

Related changes:

- Add error logging for saveApplication mutation and remove raw error details from the alert
- Bug fix for – User was unable to remove the last Planned Publication. This bug was not logged.

Unrelated changes:

- Remove unused files
  - config/ClinicalDataConfig.tsx
  - config/PreCancerTypesConfig.ts
  - useConditionalWrapper.tsx (Unused after #27)

### Related Ticket(s)

CRDCDH-1861 (FE Task)
CRDCDH-1846 (User Story)


src/content/questionnaire/sections/B.tsx
```diff
@@ -128,6 +120,13 @@ const FormSectionB: FC<FormSectionProps> = ({ SectionOption, refs }: FormSection
           : "",
       })) || [];
 
+    combinedData.study = {
+      // Restore dbGaPID from Section C
```
Alejandro-Vega, line 124: Correct me if I'm understanding wrong here, but I think we only have to do this because while creating the `combinedData` object the formObject is replacing the entire `formObject.study` instead of replacing just the properties that exist within formObject that are nested. Could we maybe just use a lodash merge for this to fix it? Just a suggestion, not a requirement. Your implementation works, I was just initially confused why we had to restore the properties.

Ex.
`const combinedData = merge(cloneDeep(data), formObject);`

Alejandro-Vega (changes requested):
Everything looks solid to me. I left one optional suggestion, but feel free to disregard it if it doesn't work or if you're fine with your existing implementation.

src/content/questionnaire/sections/B.tsx
amattu2: Thanks for this suggestion. I tested a bunch of different scenarios and Lodash merge seems to work as expected. I've migrated both B/C to this util in 5f47162. I wasn't sure if this should be applied to A/D as well, so let me know and I can update those too.

Unrelated to this, but I wanted to point out this change (since I made it last minute) – On Section B, I discovered it was impossible to remove the last Planned Publication from the form (it would always come back). I applied the same change we use for any other array-based field here 50ed153.

Alejandro-Vega: It should likely be applied to A/D as well, but since they aren't currently causing issues, let's just table that for the future as it would require quite a lot of additional testing for no additional benefit. 

I tested out the Planned Publication issue you mentioned. I do have one potential solution, but it may cause additional issues. I think your solution is probably the safest choice, so we will go with that. 

If you agree, I'll let you resolve this discussion. Otherwise, let me know if you would like to go a different direction.

Alejandro-Vega (approved):
Pre-approving, everything LGTM! Feel free to merge once you resolve the discussion above.

src/content/questionnaire/sections/B.tsx
amattu2: Got it. If we get any significant changes in A or D, we can probably implement it then.

As for the Planned Pubs issue – Since it's the most tested solution (we use it multiple times in other places), let's go with it for now. If we're ever given a chance, I'd be my preference to just rewrite the entire Submission Request functionality to use React Hook Form, which would probably solve most of the quirks/work-arounds we have to use. 

Resolving this for now.

PR #542, opened 2024-11-21
CRDCDH-2064 Remove ORCID input from Submission Request form

Alejandro-Vega:
### Overview

Temporarily removing ORCID input from Submission Request form, will be re-enabled in 3.2.0.

- ORCID input was removed, but the value for the ORCID property sent to the BE remains an empty string.

> [!Note]
> Just noticed we never added ORCID to the Review page. We should add that to the requirements when re-enabling in 3.2.0.

### Related Ticket(s)

[CRDCDH-2064](https://tracker.nci.nih.gov/browse/CRDCDH-2064) (User Story)


amattu2 (approved):
LGTM.

Agreed with adding it to the review page, also.

PR #543, opened 2024-11-21
CRDCDH-2064 Fix detecting change in ORCID for legacy forms 

Alejandro-Vega:
### Overview

Updated `questionnaireData` on API retrieval to be an empty string as well as the ORCID value in the form to be an empty string. This way it will treat ORCID property as unchanged and ignore the field.

N/A

### Related Ticket(s)

[CRDCDH-2064](https://tracker.nci.nih.gov/browse/CRDCDH-2064) (User Story)


amattu2 (approved):
LGTM! Thanks for the quick fix.

PR #550, opened 2024-11-27
CRDCDH-2067 Prevent invalid collaborators from being re-added

Alejandro-Vega:
### Overview

When an invalid collaborator is removed, it shouldn't allow the user to re-add them. Therefore, the maxCollaborators logic was updated to take this into account to prevent showing an empty dropdown.

- Kept track of all 'unavailable' or invalid collaborators
- Used that as well as the ones currently displaying as a row. Once the invalid collaborator row is removed, it will lower the maxCollaborators count as well to prevent adding that row back

### Related Ticket(s)

[CRDCDH-2067](https://tracker.nci.nih.gov/browse/CRDCDH-2067) (Bug)


amattu2 (approved):
LGTM and seems to work as expected.

PR #559, opened 2024-12-09
CRDCDH-2115 Fix Multiple Submit of Create Submission Form

amattu2:
### Overview

This PR attempts to resolve the ability for users to create >1 Data Submissions by rapidly clicking the Create button or pressing enter multiple times quickly.

React hook form already has functionality to prevent multiple concurrent form submissions, but it requires that the submit handler be async. 

> [!NOTE] 
> This change probably shouldn't be merged directly into 3.2.0, since I made some changes to the dialog there, I don't think this is forward-compatible.

N/A

### Related Ticket(s)

CRDCDH-2115


Alejandro-Vega (approved):
LGTM! Seems to work well. Nice job!

PR #560, opened 2024-12-10
CRDCDH-2086 Add OMB/PANS banner

amattu2:
### Overview

This PR adds a OMB banner to the Submission Request page, before the first question.

> [!NOTE]
> There is some slight styling differences between the implementation and design. In particular, the page banner background cuts off much higher up, rather than fading out. Since it's not directly related to this change, I left it untouched. 
>
> I also noticed that the sidebar (ProgressBar) no longer follows the top of the screen when scrolling, even though the styling for it has it set to `position: sticky`. Unsure if this is expected or when it broke, but I left it untouched.

> [!NOTE]
> I wanted to limit the changes since we're close to the 3.1.0 release, but in the future, I think we should migrate the FormContainer "Return to All Submissions" to a prefixElement instead (defined within the Review page). Just something to consider.

- Add OMB (official name is Privacy Act Notification Statement) banner to the Submission Request Section A
- Add related tests for testable requirements (approval number and expiration date)

### Related Ticket(s)

CRDCDH-2086 (FE Task)
CRDCDH-2085 (Design, Option 5)
CRDCDH-2068 (US)

### DIFF
--- a/src/components/PansBanner/index.test.tsx
+++ b/src/components/PansBanner/index.test.tsx
@@ -1,0 +1,40 @@
+import { render } from "@testing-library/react";
+import { axe } from "jest-axe";
+import dayjs from "dayjs";
+import PansBanner from "./index";
+
+describe("Accessibility", () => {
+  it("should have no violations", async () => {
+    const { container } = render(<PansBanner />);
+
+    expect(await axe(container)).toHaveNoViolations();
+  });
+});
+
+describe("Basic Functionality", () => {
+  it("should render the PANS banner without crashing", () => {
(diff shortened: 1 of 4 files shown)


Alejandro-Vega (commented):
Won't be addressed in this PR, but regarding the sticky sidebar no longer working, it seems to have been broken for a very long time. Unsure if removing the overflow causes issues else-where, but it seems to be the cause.

```javascript
// src/layouts/index.tsx

const StyledWrapper = styled("main")({
  minHeight: "400px",
  overflowX: "hidden", // Removing this fixes the "position: sticky" sidebar
});
```

PR #562, opened 2024-12-11
CRDCDH-2008 Submission Request – Dynamic Population of Program List

Alejandro-Vega:
### Overview

Updated Submission Request form to support providing a dynamic Program list in the dropdown using the listOrganizations API.

> [!NOTE]
> Currently the BE doesn't support adding the Program to Org collection after the Submission is approved. 

- Replaced hard-coded program list with dynamic list from listOrganizations API
- Updated Program logic to better fit with the new dynamic list
- Removed utility functions that were no longer needed regarding Programs
- Added `programAbbreviation` and `programDescription` to the root-level of the Application
- Updated default values and tests

### Related Ticket(s)

[CRDCDH-2008](https://tracker.nci.nih.gov/browse/CRDCDH-2008) (Task)
[CRDCDH-1901](https://tracker.nci.nih.gov/browse/CRDCDH-1901) (US)


src/content/questionnaire/sections/B.tsx
```diff
@@ -314,8 +292,12 @@ const FormSectionB: FC<FormSectionProps> = ({ SectionOption, refs }: FormSection
         <SelectInput
           id="section-b-program"
           label="Program"
-          options={programOptions?.map(programToSelectOption)}
-          value={programOption?.name}
+          name="program[_id]"
```
amattu2, line 295: I noticed two issues with the selection field, I think they're both related:

1) If you select the "Other" option, and view the form in Read Only mode (e.g. Federal Lead or if you submit), when you try to reload the page, you get the unsaved changes alert and the field is outlined in red.


Screenshot 2024-12-12 at 10 09 32 AM

---

2) When you select the "Other" option, then move on to the next sections, the "Other" no longer persists in the dropdown. I like that you cleaned up all the other unnecessary properties stored in the program object, so without adding those back, do you think there's a way we can safely infer if the Other option was selected?

Screenshot 2024-12-12 at 10 17 19 AM

Screenshot 2024-12-12 at 10 18 18 AM

src/content/questionnaire/Controller.tsx
```diff
@@ -15,11 +16,13 @@ const QuestionnaireController = () => {
 
   if (appId) {
     return (
-      <InstitutionProvider>
-        <FormProvider id={appId}>
-          <FormView section={section} />
-        </FormProvider>
-      </InstitutionProvider>
+      <OrganizationProvider preload>
```
amattu2, line 19: The current OrganizationProvider doesn't query for the Organization Abbreviation or Descriptions (since it isn't used anywhere else). Could you add these fields to the listOrganizations query? It currently isn't populating those two fields in the SR.

src/content/questionnaire/Controller.tsx
```diff
@@ -3,6 +3,7 @@ import FormView from "./FormView";
 import ListView from "./ListView";
 import { FormProvider } from "../../components/Contexts/FormContext";
 import { InstitutionProvider } from "../../components/Contexts/InstitutionListContext";
+import { OrganizationProvider } from "../../components/Contexts/OrganizationListContext";
```
amattu2, line 6: No action needed here – In the future, I'm wondering if we should Memoize the InstitutionProvider and OrganizationProvider so it doesn't refetch each time we open a submission request?

Unsure if that would cause issues, since either of those two providers may have updated data at any point. Just something to consider down the road.

src/content/questionnaire/sections/B.tsx
```diff
@@ -340,7 +322,7 @@ const FormSectionB: FC<FormSectionProps> = ({ SectionOption, refs }: FormSection
           id="section-b-program-abbreviation"
           label="Program Abbreviation"
           name="program[abbreviation]"
-          value={predefinedProgram ? programOption?.abbreviation : program?.abbreviation}
+          value={program?.abbreviation}
```
amattu2, line 325: No action needed here – But the abbreviation field is filtering most special characters out from the pre-selected options. Do you recall if that was a requirement from somewhere? I didn't add any character restrictions on the Edit Program > Abbreviation field, I'm guessing I need to revisit that to ensure it's consistent.

amattu2 (changes requested):
Nice job. Noticed two issues with the Program dropdown, and I annotated 2 questions that don't require any action.

src/content/questionnaire/Controller.tsx
Alejandro-Vega: Good catch, added in [9c981de](https://github.com/CBIIT/crdc-datahub-ui/pull/562/commits/9c981dec2708b5b483d8ae0fe1efef22311f8a5f).

src/content/questionnaire/sections/B.tsx
Alejandro-Vega: Thanks for catching this. This was a big oversight by me. My new implementation was not backwards compatible due to the removing of properties. I have added support for older forms in [9cbbfc3](https://github.com/CBIIT/crdc-datahub-ui/pull/562/commits/9cbbfc3074b4da1785ab0499a34ba48aa48dc706).

In order to differentiate from an empty form with "Other" vs "Not Applicable", I had to add a check for the existence of the old property "notApplicable". This was the only way to fully support older forms. LMK if you can think of a better way.

Please re-test and verify this for me as there are a lot of possible scenarios in this form.

Alejandro-Vega: Additional note, for some reason if you try saving a brand new, un-edited, submission then the select would wanna default to the first option in the select, in this case being the NotApplicableProgram. Fixed this by adding an empty option in the ProxySelect. [cff287b](https://github.com/CBIIT/crdc-datahub-ui/pull/562/commits/cff287bd4a965fcfe0dc9a54f23571071655c4c0).

src/content/questionnaire/Controller.tsx
Alejandro-Vega: Agreed, will address in a separate PR in the future. This will require more testing.

src/content/questionnaire/sections/B.tsx
Alejandro-Vega: Yes, this comes from [CRDCDH-35](https://tracker.nci.nih.gov/browse/CRDCDH-35) "Validation: Only alpha-numerical, ‘-‘ and space are allowed. ". Back in the double digit ticket days 😢.

src/content/questionnaire/sections/B.tsx
Alejandro-Vega: Will let you resolve this if you feel it resolved the issues you described.

src/content/questionnaire/sections/B.tsx
amattu2: Thanks for tracking that down. I'll update the Edit Program page to match this

amattu2 (approved):
Changes LGTM and it seems to be consistent with the current DEV2 implementation.

PR #564, opened 2024-12-16
CRDCDH-2113 Create a Data Submission – PBAC Studies Support

amattu2:
### Overview

This PR is preparation for Data Submission PBAC controls. Instead of explicitly utilizing the account's assigned studies, we now determine if the Create a Data Submission dialog should utilize the AuthCtx studies or the full list. 

- Cleanup unused mock queries for list orgs in the Create a DS dialog tests
- Define `RequiresStudiesAssigned` variable, which defines the roles that enforce a study selection
- Update Create a DS dialog to utilize the listApprovedStudies instead of their assigned studies based on:
  - Is the dialog even open?
  - Does the user role require studies be assigned? If not, we call the API
  - Does the account have the "All" study assigned? If yes, we call the API

### Related Ticket(s)

CRDCDH-2113 (Task)
CRDCDH-2023 (US 1 of 2)
CRDCDH-1886 (US 2 of 2)

### DIFF
--- a/src/components/DataSubmissions/CreateDataSubmissionDialog.test.tsx
+++ b/src/components/DataSubmissions/CreateDataSubmissionDialog.test.tsx
@@ -10,13 +10,7 @@ import {
   ContextState as AuthCtxState,
   Status as AuthStatus,
 } from "../Contexts/AuthContext";
-import {
-  CREATE_SUBMISSION,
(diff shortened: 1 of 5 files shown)


Alejandro-Vega (approved):
LGTM! Works well!

PR #566, opened 2024-12-17
CRDCDH-2123 Support "All" studies for Federal Leads

amattu2:
### Overview

This PR adds the "All" studies option for Federal Leads on the Edit User tool.

> [!Warning]
> When switching roles from anything (e.g. `Submitter`) to `Federal Lead`, MUI throws a warning because the "All" option is programmatically selected, but not available in the study options during the first render. The only solution I could find that worked is [not very clean](https://github.com/CBIIT/crdc-datahub-ui/pull/566/commits/f6e0da767afaddc2178114a8545b510b2ddcb373). I'm open to any alternatives.

- Add `All` option to the Studies list for Federal Leads, selected by default
- Migrate `VisibleFieldState` out of ProfileView into the useProfileFields hook

### Related Ticket(s)

CRDCDH-2123 (Task)

### DIFF
--- a/src/content/users/ProfileView.tsx
+++ b/src/content/users/ProfileView.tsx
@@ -302,10 +302,10 @@ const ProfileView: FC<Props> = ({ _id, viewType }: Props) => {
   ) => {
     let updatedData = [...data];
 
-    // Previous studies included all but the user selected other studies
+    // Previous studies included "All", but the user selected something different
     if (prevStudiesRef.current?.includes(ALL_STUDIES_OPTION)) {
       updatedData = updatedData.filter((v) => v !== ALL_STUDIES_OPTION);
-      // User selected all studies, remove all other studies
+      // User selected "All" studies, remove any other studies
     } else if (data.includes(ALL_STUDIES_OPTION)) {
       updatedData = [ALL_STUDIES_OPTION];
     }
--- a/src/content/users/ProfileView.tsx
+++ b/src/content/users/ProfileView.tsx
@@ -2,7 +2,7 @@ import { FC, useEffect, useMemo, useRef, useState } from "react";
 import { useLazyQuery, useMutation, useQuery } from "@apollo/client";
 import { LoadingButton } from "@mui/lab";
 import { Box, Container, MenuItem, Stack, TextField, Typography, styled } from "@mui/material";
-import { Controller, ControllerRenderProps, useForm } from "react-hook-form";
+import { Controller, ControllerRenderProps, SubmitHandler, useForm } from "react-hook-form";
 import { useNavigate } from "react-router-dom";
 import { useSnackbar } from "notistack";
 import bannerSvg from "../../assets/banner/profile_banner.png";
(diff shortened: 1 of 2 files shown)


amattu2:
Coveralls is rejecting this PR since it has a significant coverage drop (not really seeing why). Regardless, that can be ignored.

src/content/users/ProfileView.tsx
```diff
@@ -460,9 +496,14 @@ const ProfileView: FC<Props> = ({ _id, viewType }: Props) => {
...
-                        options={studyOptions}
+                        options={
+                          roleField === "Federal Lead" &&
+                          !studyOptions?.includes(ALL_STUDIES_OPTION)
+                            ? [ALL_STUDIES_OPTION, ...studyOptions]
+                            : studyOptions
+                        }
                         getOptionLabel={(option: string) => formattedStudyMap[option]}
-                        onChange={(_, data: string[]) => field.onChange(data)}
+                        onChange={(_, data: string[]) => handleStudyChange(field, data)}
```
Alejandro-Vega, line 506: Unsure of where exactly to put this, but related to your issue where you switch from another role to "Federal Lead". After you switch from another role and click the studies dropdown, the first click doesn't actually select any new studies. After the first click the rest work as expected. Even with my fix suggestion this doesn't go away. If this is a MUI thing, then disregard.

src/content/users/ProfileView.tsx
```diff
@@ -460,9 +496,14 @@ const ProfileView: FC<Props> = ({ _id, viewType }: Props) => {
                             </StyledTag>
                           );
                         }}
-                        options={studyOptions}
+                        options={
```
Alejandro-Vega, line 499: I do have one potential suggestion to fix your issue when switching from other roles to "Federal Lead". It seems to be a timing issue, Instead of assigning the "All" option at the time of role change, you can instead do it after sorting the study options which happens after every role change anyways due to the addition of the "All" option.

This was just a rough suggestion, if you find any issues with it feel free not to use it.

```javascript
const handleRoleChange = (field: ControllerRenderProps<FormInput, "role">, value: UserRole) => {
    if (prevRoleRef.current === "Federal Lead") {
      setValue(
        "studies",
        studiesField.filter((v) => v !== ALL_STUDIES_OPTION)
      );
    } else if (value === "Federal Lead") {
      setValue("studies", []); // updated line
    }

    field.onChange(value);
  };
  
  ...

useEffect(() => {
    if (fieldset.studies !== "UNLOCKED") {
      return;
    }
    sortStudyOptions();

    if (roleField === "Federal Lead") {
      setValue("studies", [ALL_STUDIES_OPTION]);
    }
  }, [formattedStudyMap, roleField]);
  
  ...
  
  <StyledAutocomplete
     ...
     options={studyOptions}
     ...
```

Alejandro-Vega (changes requested):
Everything LGTM and works great! I left a potential fix for the timing issue regarding the "All" option below. As well as found one potential bug or just a MUI thing. Feel free to disregard both if needed.

src/content/users/ProfileView.tsx
amattu2: Thanks for looking into this. Your solution seems to work perfectly and it's much cleaner. Updated in 9173800.

amattu2: Applied that change in 7f729d5. I rechecked all of the different scenarios and everything seems to work as expected now.

LMK if you find anything else.

Alejandro-Vega (approved):
LGTM! Working well with updated changes.

PR #572, opened 2024-12-27
CRDCDH-2105 PBAC – Individual Permissions Control of User Accounts

amattu2:
### Overview

This PR introduces the ability to configure individual PBAC permissions for an account.

> [!Warning]
> Some known issues / expected things from the backend implementation:
> - You can adjust a Submitter or User's permissions, but no changes will actually be saved. This is expected, but BE should be locking all permissions, and they currently aren't
> - The backend seems to reject some permissions and/or notification selections and returns an error for it. Just uncheck them in the FE for now
> - The visual permission names don't really match the user stories anymore (they keep getting changed)
> - The order of actual permissions in the nested groups may not match the user stories (see below)

> [!NOTE]
> The frontend sorts the groups based on the User Story, but we don't sort the nested permissions themselves since that's somewhat impractical (way too many options). 

- Implement a PermissionsPanel component, which handles the PBAC checkboxes, defaults, and form management
- Update the useProfileFields hook to conditionally render permissions/notifications fields
- Update remaining queries to pull the notification/permission assignments
- Update the Apollo memory cache to stop it from normalizing the PBAC data
  - It was discarding the relationship between the roles and their specific defaults causing serious issues
- Move `AuthPermissions` to a separate file (`PBAC.d.ts`) and create `AuthNotifications` type for consistency
  - I'm also fine with just leaving it in `Auth.d.ts` if that's preferred, it was just getting cluttered
- Update the ProfileView action buttons to be consistent with design (matching DS and SR theme)

### Related Ticket(s)

CRDCDH-2105 (Task)
... A lot of stories

### DIFF
--- a/src/components/PermissionPanel/index.tsx
+++ b/src/components/PermissionPanel/index.tsx
@@ -86,23 +86,12 @@ const StyledNotice = styled(Typography)({
   userSelect: "none",
 });
 
-type PermissionPanelProps = {
-  /**
-   * The original/stored role of the user.
-   *
-   * This is used to determine if the role has changed and to update the default permissions.
-   *
-   * @deprecated This prop is no longer used. Remove it.
-   */
-  role: UserRole;
-};
-
 /**
  * Provides a panel for managing permissions and notifications for a user role.
  *
  * @returns The PermissionPanel component.
  */
-const PermissionPanel: FC<PermissionPanelProps> = ({ role }) => {
+const PermissionPanel: FC = () => {
   const { enqueueSnackbar } = useSnackbar();
   const { setValue, watch } = useFormContext<EditUserInput>();
 
--- a/src/components/PermissionPanel/index.tsx
+++ b/src/components/PermissionPanel/index.tsx
@@ -1,5 +1,5 @@
 import { useQuery } from "@apollo/client";
-import { ArrowDropDown } from "@mui/icons-material";
+import ExpandMoreIcon from "@mui/icons-material/ExpandMore";
 import {
   Accordion,
   AccordionDetails,
@@ -10,10 +10,10 @@ import {
   FormGroup,
   styled,
   Typography,
+  Unstable_Grid2 as Grid2,
 } from "@mui/material";
 import { FC, memo, useEffect, useMemo, useRef } from "react";
 import { useFormContext } from "react-hook-form";
-import Grid2 from "@mui/material/Unstable_Grid2";
 import { useSnackbar } from "notistack";
 import { cloneDeep } from "lodash";
 import {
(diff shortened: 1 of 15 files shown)


src/utils/profileUtils.ts
```diff
@@ -40,3 +40,103 @@ export const userToCollaborator = (
...
+   * The PBACDefaults for the group
+   */
+  data: PBACDefault<T>[];
+}[][];
+
+/**
+ * A utility function to group an array of PBACDefaults into columns
+ * based on the group name.
+ *
+ * If colCount is greater than the number of groups, the last column will
```
Alejandro-Vega, line 64: I could be misunderstanding, but I believe it should be reversed since you are looping through the groups and checking if that is greater. Ex. "If number of groups is greater than colCount, ...".

Although, your implementation does seem correct and working as expected.

src/components/PermissionPanel/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+const StyledAccordionSummary = styled(AccordionSummary)({
+  borderBottom: "1px solid #6B7294",
+  minHeight: "unset !important",
+  "& .MuiAccordionSummary-content": {
+    margin: "9px 0",
+  },
+  "& .MuiAccordionSummary-content.Mui-expanded": {
+    margin: "9px 0",
+  },
+  "& .MuiAccordionSummary-expandIcon": {
```
Alejandro-Vega, line 47: This is not correctly targeting the expand icon, I still see it as gray. I believe it should be `.MuiAccordionSummary-expandIconWrapper` instead.

src/components/PermissionPanel/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+  };
+
+  useEffect(() => {
+    handleRoleChange(selectedRole);
+  }, [selectedRole]);
+
+  return (
+    <StyledBox>
+      <StyledAccordion elevation={0} data-testid="permissions-accordion">
+        <StyledAccordionSummary expandIcon={<ArrowDropDown />}>
```
Alejandro-Vega, line 200: Just a suggestion, I think MUI has an icon that is closer to the design. Feel free to switch if you'd like.
`import ExpandMoreIcon from "@mui/icons-material/ExpandMore";`

Alejandro-Vega (changes requested):
Code looks solid, and I like the changes you made with separating the PBAC types to their own file. I left a few super small and optional suggestions. Feel free to disregard them.

src/utils/profileUtils.ts
amattu2: You're correct. I wrote the description then didn't proof read it. Let me know if this makes more sense:

>  * If the number of unique groups exceeds `colCount`, the function will
>  * aggregate the remaining groups into the last column.

Updated in 99a161f.

src/components/PermissionPanel/index.tsx
amattu2: Good catch. Updated in 4553245.

src/components/PermissionPanel/index.tsx
amattu2: Agreed. Updated with the color fix 4553245.

src/utils/profileUtils.ts
Alejandro-Vega: LGTM!

PR #574, opened 2024-12-30
CRDCDH-2118 Operation Dashboard – Studies filter utilizes "All" studies selected

Alejandro-Vega:
### Overview

Cleaned up Operation Dashboard logic for content parameters. Also, skipped adding studies parameter when user has "All" study, so that it shows all studies. 

- Separated some of the logic for adding content params to the dashboard to utility functions with tests
- Skipped adding studies parameter when user has "All" study assigned

### Related Ticket(s)

[CRDCDH-2118](https://tracker.nci.nih.gov/browse/CRDCDH-2118) (Task)
[CRDCDH-2155](https://tracker.nci.nih.gov/browse/CRDCDH-2155) (US 1)
[CRDCDH-1886](https://tracker.nci.nih.gov/browse/CRDCDH-1886) (US 2)
[CRDCDH-1887](https://tracker.nci.nih.gov/browse/CRDCDH-1887) (US 3)

### DIFF
--- a/src/utils/dashboardUtils.ts
+++ b/src/utils/dashboardUtils.ts
@@ -2,8 +2,12 @@ import { DashboardContentOptions } from "amazon-quicksight-embedding-sdk";
 import { Logger } from "./logger";
 
 /**
- * A helper that decides whether to add the "studiesParameter" for
- * roles that require assigned studies (e.g., Federal Lead).
+ * Constructs and returns an array of QuickSight parameter objects for a user's studies.
+ *
+ * - If the user's first study is `All`, the function returns an empty array (allowing QuickSight to display all data).
+ * - If the user has a valid array of studies, it creates a `studiesParameter` whose values are the `_id` fields of each study.
+ * - Otherwise, it logs an error and returns a parameter array with `["NO-CONTENT"]`.
+ *
  *
  * @param {User} user - The current user
  * @returns {DashboardContentOptions["parameters"]} The updated dashboard parameters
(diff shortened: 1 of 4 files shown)


src/utils/dashboardUtils.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param {User} user - The current user
+ * @returns {DashboardContentOptions["parameters"]} The updated dashboard parameters
+ */
+export const addStudiesParameter = (user: User): DashboardContentOptions["parameters"] => {
+  const params: DashboardContentOptions["parameters"] = [];
+  const { studies } = user || {};
+
+  // If "All" is the first study, do NOT push the "studiesParameter"
+  if (studies?.[0]?._id === "All") {
```
amattu2, line 20: Technically this assertion is correct, the "All" study should be the only one assigned and in the first index. However, I suspect there may be instances where this isn't true.

e.g. If a Federal Lead has permission to create a Submission Request, if that request is approved, I suspect BE would assign the newly created study to the Federal Lead's profile. If the study happens to be inserted at the first spot, that would break this logic.

The chances of these sort of issues happening is slim, but if you agree that it could happen, maybe let's just search the entire array of studies?

I did something like this in the DS Creation dialog, feel free to adapt:

```javascript
(studies || [])?.findIndex((s) => s?._id === "All") !== -1
```

amattu2 (changes requested):
Tested locally and works as expected. I noted one edge case / scenario that we may want to address, but it's up to you.

amattu2 (approved):
LGTM!

PR #576, opened 2025-01-02
CRDCDH-2021 Collaborators Dialog Updates

Alejandro-Vega:
### Overview

Removed Organization and Access columns from Collaborators dialog table.

> [!NOTE]
> The Collaborators dialog needs an updated description. Will update in future PR if still have not received updated text yet.

> [!NOTE]
> Haven't tested with BE yet as it is not ready yet, but based my implementation on the existing BE PR.

- Removed the Organization and Access columns from Collaborators dialog table
- All collaborators now always have a "Can Edit" permission
- Updated Collaborator name to match BE formatting. (Otherwise it shows differently before and after saving a collaborator) (See https://github.com/CBIIT/crdc-datahub-backend/pull/527/files)
- Removed any mention of user org and "Can View" permission
- Looks like the org property still exists in Collaborator type on BE, just defaulted to null and removed it from being returned as it is not used on FE
- Removed org property from the User type
- Increased max characters before truncation from 10 => 35 due to extra space. Added comment in ticket. 

### Related Ticket(s)

[CRDCDH-2021](https://tracker.nci.nih.gov/browse/CRDCDH-2021) (Task)
[CRDCDH-1984](https://tracker.nci.nih.gov/browse/CRDCDH-1984) (US)


src/utils/profileUtils.ts
```diff
@@ -1,5 +1,3 @@
-import { formatName } from "./stringUtils";
```
amattu2, line 1: I think the `formatName` util is unused now? If we have a future use for it, I'm fine with keeping it, otherwise let's remove it too.

src/types/Submissions.d.ts
```diff
@@ -373,10 +373,10 @@ type SubmitButtonResult = {
 /**
  * Represents the permissions a collaborator can have in a submission
  */
-type CollaboratorPermissions = "Can View" | "Can Edit";
+type CollaboratorPermissions = "Can Edit";
 
 /**
- * Represents a submitter that can view/edit another submitter's submission
+ * Represents a submitter that can view and edit another submitter's submission
```
amattu2, line 379: Do we need to remove `Organization` from the `Collaborator` type too?

amattu2 (changes requested):
Everything generally LGTM. Can't fully test locally since the API isn't updated yet. I commented two super minor things.

BTW, the US was updated with the new dialog content, so we can probably include that in this PR too?

src/utils/profileUtils.ts
Alejandro-Vega: Agreed, removed in [68cf98e](https://github.com/CBIIT/crdc-datahub-ui/pull/576/commits/68cf98e151a84aef6447d2135cb921881371cf21).

src/types/Submissions.d.ts
Alejandro-Vega: Last I checked BE kept the Org in the type, but we don't currently have a use for it. I removed in [17de629](https://github.com/CBIIT/crdc-datahub-ui/pull/576/commits/17de629ebdecc8c36bb2ce983cf6c4b7d09972ad) and [edef527](https://github.com/CBIIT/crdc-datahub-ui/pull/576/commits/edef5276f602f8677480ea2a54d6cf6a6ed7cf81).

Alejandro-Vega:
> Everything generally LGTM. Can't fully test locally since the API isn't updated yet. I commented two super minor things.
> 
> BTW, the US was updated with the new dialog content, so we can probably include that in this PR too?

Updated to include new dialog description text in [a1d0d61](https://github.com/CBIIT/crdc-datahub-ui/pull/576/commits/a1d0d610992168f51e5ae75a087b245d3ac20cf6).

src/types/Submissions.d.ts
amattu2: Thanks @Alejandro-Vega. Since we don't have a use for it, I think it makes sense to remove it. Likely, BE will remove it at some point.

One last place I noticed it was queried for:

https://github.com/CBIIT/crdc-datahub-ui/blob/a1d0d610992168f51e5ae75a087b245d3ac20cf6/src/graphql/getSubmission.ts#L65-L68

Alejandro-Vega: Good catch, removed in [eb122f3](https://github.com/CBIIT/crdc-datahub-ui/pull/576/commits/eb122f357488162fdf5019916213ed179a431609).

amattu2 (approved):
LGTM!

Didn't retest changes locally.

PR #577, opened 2025-01-02
CRDCDH-2102 fix: The `useProfileFields` hook should utilize PBAC

amattu2:
### Overview

This PR updates the `useProfileFields` hook to utilize PBAC instead of RBAC when determining which fields are unlocked, disabled, etc.

> [!NOTE]
> I can't really check this change since the backend won't return data for non-Admins ATM.

N/A

### Related Ticket(s)

CRDCDH-2102 (FE Task)


Alejandro-Vega (approved):
LGTM! Didn't test locally.

PR #578, opened 2025-01-02
CRDCDH-1674 Fix Orphaned Files error being moved from submission to QC Results

Alejandro-Vega:
### Overview

The BE moved the Orphaned Files error from `Submission.fileErrors` to within the QC Results API. Updated submit button config to accept QC Results as a param when checking logic to avoid submitting and admin submit when there's orphaned files.

- Added calling QC Results API in parallel to getSubmission API within SubmissionContext, including while polling
- Added the QC Results to the Submit button config to use for determining whether or not to enable the Submit button
- Updated the Orphaned Files logic to use QC Results instead of `Submission.fileErrors`

### Related Ticket(s)

[CRDCDH-1674](https://tracker.nci.nih.gov/browse/CRDCDH-1674) (Bug)


amattu2 (approved):
LGTM and seems to work as expected.

----

Thanks for updating the implementation, although this is insanely inefficient now IMO. I think we will need the BE to add a flag on the submission object in the future or something.

PR #580, opened 2025-01-03
HOTFIX: Create a Data Submission List View Error

amattu2:
### Overview

This is a minor fix for the generic error `"An error occurred while loading the data."` that appears when creating data submissions using a study that isn't associated with any organization (program).

It can't be replicated all the time, but it seems to do it with the study `1901-2`

- Add null-safe chaining when sorting the Program List for the dropdown filter
- Add the logging util to the try-catch

### Related Ticket(s)

N/A


Alejandro-Vega (approved):
Couldn't replicate myself locally, but code LGTM!

PR #581, opened 2025-01-03
CRDCDH-2186 Allow save without changes

amattu2:
### Overview

Previously, we prevented unnecessary API calls by comparing the saved SR with the current state of the SR when pressing the save button; In order to support upcoming auto-delete Submission Request functionality changes, we must always dispatch the saveApplication API call when the user requests it. 

> [!NOTE]
> This is based on the PBAC branch to keep merging simple, but it's very unrelated to the PBAC changes.

- Remove `isEqual` comparison on saveApp functionality

### Related Ticket(s)

CRDCDH-2186 (FE Task)


Alejandro-Vega (approved):
LGTM!

PR #582, opened 2025-01-03
Sync pbac-update-user-roles with 3.2.0

amattu2:
### Overview

Doing this via a PR since it required several manual resolutions.


Alejandro-Vega (approved):
LGTM, I don't see any issues.

PR #586, opened 2025-01-08
fix: CLI Tool should not be bound by `data_submission:create`

amattu2:
### Overview

PR to restore the original visibility of the Uploader CLI Tool, which should always be visible and not hidden based on PBAC. Also, moves the CLI Tool to right after User Profile.

N/A

### Related Ticket(s)

N/A – PBAC Stories


Alejandro-Vega (approved):
LGTM!

PR #587, opened 2025-01-08
CRDCDH-2212 Restrict Data Submission 'create' permission scope for specific roles

Alejandro-Vega:
### Overview

The Data Submission create permission for 'Federal Lead' and 'Data Commons Personnel' roles were updated to restrict it to only applying to their own Data Submission.

- Updated 'Federal Lead' role Data Submission create permission to only be allowed to apply to their own Data Submissions
- Updated 'Data Commons Personnel' role Data Submission create permission to only be allowed to apply to their own Data Submissions
- Updated all Data Submission permissions to accept "All" study for Federal Leads as a valid study

### Related Ticket(s)

[CRDCDH-2212](https://tracker.nci.nih.gov/browse/CRDCDH-2212) (Task)
[CRDCDH-2023](https://tracker.nci.nih.gov/browse/CRDCDH-2023) (US 1)
[CRDCDH-1886](https://tracker.nci.nih.gov/browse/CRDCDH-1886) (US 2)
[CRDCDH-1887](https://tracker.nci.nih.gov/browse/CRDCDH-1887) (US 3)

### DIFF
--- a/src/config/AuthPermissions.ts
+++ b/src/config/AuthPermissions.ts
@@ -81,10 +81,10 @@ export const PERMISSION_MAP = {
       if (role === "Submitter" && isSubmissionOwner && hasPermissionKey) {
         return true;
       }
-      if (role === "Federal Lead" && hasPermissionKey) {
+      if (role === "Federal Lead" && isSubmissionOwner && hasPermissionKey) {
         return studies?.some((s) => s._id === submission.studyID);
       }
-      if (role === "Data Commons Personnel" && hasPermissionKey) {
+      if (role === "Data Commons Personnel" && isSubmissionOwner && hasPermissionKey) {
         return dataCommons?.some((dc) => dc === submission?.dataCommons);
       }
       if (role === "Admin" && hasPermissionKey) {
--- a/src/config/AuthPermissions.ts
+++ b/src/config/AuthPermissions.ts
@@ -82,7 +82,7 @@ export const PERMISSION_MAP = {
         return true;
       }
       if (role === "Federal Lead" && isSubmissionOwner && hasPermissionKey) {
-        return studies?.some((s) => s._id === submission.studyID);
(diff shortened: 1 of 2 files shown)


Alejandro-Vega:
@amattu2  Let me know if I missed any other requirement changes. These are the only changes I noticed.

src/config/AuthPermissions.ts
```diff
@@ -81,10 +81,10 @@ export const PERMISSION_MAP = {
       if (role === "Submitter" && isSubmissionOwner && hasPermissionKey) {
         return true;
       }
-      if (role === "Federal Lead" && hasPermissionKey) {
-        return studies?.some((s) => s._id === submission.studyID);
+      if (role === "Federal Lead" && isSubmissionOwner && hasPermissionKey) {
+        return studies?.some((s) => s._id === submission.studyID || s._id === "All");
```
amattu2, line 85: If I'm thinking about this correctly, and it's possible I'm not, both Federal Leads and Data Commons Personnel don't need the additional check against their studies/data commons for the create permission. 

I think we would then just simplify the PBAC check for create to something like:

```
// ... existing check for collaborators

// Remove any checking of roles, since no roles have specific conditions now?
if (isSubmissionOwner && hasPermissionKey) {
  return true;
}

// base case, fallback to false
return false;
```

Right now, if a **ICDC** (for example) Data Commons Personnel creates a Data Submission for **CDS**, they effectively wouldn't be able to interact with it.

Alejandro-Vega: Makes sense, updated in [953ef63](https://github.com/CBIIT/crdc-datahub-ui/pull/587/commits/953ef63805390e158fb50f28d02e701c186b4a65).

amattu2 (approved):
Changes LGTM. The action failures are also present in the base branch, so I'm just going to merge anyway.

PR #589, opened 2025-01-09
feat: Sort PBACDefaults within their groups

amattu2:
### Overview

This PR fixes an oversight where the order of the PBAC option must be sorted based on an arbitrary order, rather than the order returned by the API.

- Sort the PBACDefaults within each group by the `order` property given
- Update coverage in multiple areas to test this

### Related Ticket(s)

N/A


Alejandro-Vega (approved):
LGTM!

PR #590, opened 2025-01-09
fix: Sort the order of AuthRoles per the US

amattu2:
### Overview

Sort the User Roles array to reflect the order described in the User Story. 

N/A

### Related Ticket(s)

N/A


Alejandro-Vega (approved):
LGTM!

PR #585, opened 2025-01-07
CRDCDH-2199 Show Review and Submit in Progress Bar when submit button is visible

Alejandro-Vega:
### Overview

A Submitter and Federal Lead can submit while a Submission request is in "Submitted" or "Inquired" status. Although, the progress bar for Federal Lead still said "Review" instead of "Review and Submit". Updated logic to show correct title when submit button is visible.

### Change Details (Specifics)

N/A

### Related Ticket(s)

[CRDCDH-2199](https://tracker.nci.nih.gov/browse/CRDCDH-2199) (Bug)

PR #591, opened 2025-01-13
Migrate CRDCDH-2199 to PBAC

amattu2:
### Overview

Proposed PR to migrate #585 to Permission Based Access Control.

Some of the scenarios I tested and compared to DEV2:

- Admin with View only
- Federal Lead with Submit permission (also tested Resubmit)
- Submitter with Submit

It seems to be correct, but I'm not 100% that this is the correct migration. This also includes updates to the test coverage so we can pass permissions in the unit tests.


Alejandro-Vega (approved):
LGTM! Seems correct and behaves correctly. We can always change in the future if needed.

PR #600, opened 2025-01-22
CRDCDH-2037 Hotfix: Reset page when table view changes

Alejandro-Vega:
### Overview

Added missing dep so that the table resets the page to 0 once the table view changes within the Validation Results tab in a Data Submission.

- If you switch to a different page, then switch table view, then the page would try to persist. Updated this so that it reset back to 0 when table view changes.

### Related Ticket(s)

[CRDCDH-2037](https://tracker.nci.nih.gov/browse/CRDCDH-2037) (Task)
[CRDCDH-1889](https://tracker.nci.nih.gov/browse/CRDCDH-1889) (US)


amattu2 (approved):
LGTM

PR #602, opened 2025-01-22
CRDCDH-2187 Submission Request Cancel/Restore

amattu2:
### Overview

This PR builds out the functionality to Cancel and Restore a Submission Request, and includes supporting QoL updates for related implementations.

> [!Warning]
> Some implementation notes:
> - The backend refers to the "cancel" action API as "delete", so in the frontend I remapped that to `cancelApplication` instead of `deleteApplication` since that's the correct terminology
> - When you delete an application, it disappears from the list. Once we have the filters in place, I assume it would be visible still (if you select the Canceled status)

- Add a Cancel/Restore button for the Submission Request List
- Add associated test coverage
- Update the GenericTable to mitigate 508 errors when using an empty column header name (Instead of using a TR, switch it to a TD)

### Related Ticket(s)

CRDCDH-2187 (Task)
CRDCDH-2146 (US 1)
CRDCDH-2142 (US 2)


src/components/CancelApplicationButton/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }, [isRestoreAction, restoreApp, cancelApp, onCancel, enqueueSnackbar]);
+
+  if (!hasPermission(user, "submission_request", "cancel", application)) {
+    return null;
+  }
+
+  return (
+    <>
+      <StyledIconButton
```
Alejandro-Vega, line 132: I do think these should really have a tooltip added in the future. Don't see any requirement for it, but just a suggestion.

src/config/AuthPermissions.ts
```diff
@@ -61,6 +66,32 @@ export const PERMISSION_MAP = {
...
     },
     review: NO_CONDITIONS,
+    cancel: (user, application) => {
+      const hasPermissionKey = user?.permissions?.includes("submission_request:cancel");
+      if (!hasPermissionKey) {
+        return false;
+      }
+
+      const isFormOwner = application?.applicant?.applicantID === user?._id;
+      const canDeleteOthers: UserRole[] = ["Admin", "Federal Lead", "Data Commons Personnel"];
```
Alejandro-Vega, line 76: Should this instead be defined in the AuthRoles file?

Alejandro-Vega (changes requested):
Everything looks solid to me. Great work! Also, I agree that the submissions should likely not disappear and Canceled submissions should still be shown. 

I left 2 small optional suggestions below. Feel free to disregard.

src/components/CancelApplicationButton/index.tsx
amattu2: Agreed. For now, I just added the `title` attribute to the button since titles are generally controlled by FE and not requirements. In the future, if needed, we can migrate the title to a tooltip. 4dcc59f

src/config/AuthPermissions.ts
amattu2: Agreed. Migrated in 1338640.

amattu2:
Back to on hold. The icons will be changed soon and we'll be adding a tooltip.

Alejandro-Vega (approved):
LGTM! The new tooltips and icons look great!

PR #606, opened 2025-01-23
CRDCDH-2265 Support multi-select status selection

amattu2:
### Overview

This PR allows multi-selection of the Data Submission List > Status Filter.

> [!NOTE]
> This PR is on hold waiting for BE.

- Update the Status input to multi-select
- Change searchParams implementation for the status field to an array-based implementation (status=x1&status=x2...)
- Update the order of the option list
- Export the default form values and use that for the Data Submission List formRef instead of duplicating fields
- Remove very very old MVP-1 note

### Related Ticket(s)

CRDCDH-2265 (Task)
CRDCDH-2206 (US)

### DIFF
--- a/src/components/DataSubmissions/DataSubmissionListFilters.tsx
+++ b/src/components/DataSubmissions/DataSubmissionListFilters.tsx
@@ -1,6 +1,6 @@
 import { memo, useEffect, useRef, useState } from "react";
 import { FormControl, IconButton, MenuItem, Grid, Box, styled, Stack } from "@mui/material";
-import { debounce, isEqual } from "lodash";
+import { debounce, isEqual, sortBy } from "lodash";
 import RefreshIcon from "@mui/icons-material/Refresh";
 import { Controller, useForm } from "react-hook-form";
 import StyledSelectFormComponent from "../StyledFormComponents/StyledSelect";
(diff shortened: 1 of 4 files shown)


src/components/DataSubmissions/DataSubmissionListFilters.tsx
```diff
@@ -229,9 +233,14 @@ const DataSubmissionListFilters = ({
...
-      newSearchParams.set("status", statusFilter);
-    } else if (statusFilter === "All") {
+    if (statusFilter && statusFilter.length > 0) {
+      newSearchParams.delete("status");
+      if (!isArrayEqual(statusFilter, defaultValues.status)) {
+        statusFilter.forEach((status) => {
+          newSearchParams.append("status", status);
+        });
+      }
+    } else if (!statusFilter || !statusFilter.length) {
```
Alejandro-Vega, line 243: This can be combined `!statusFilter?.length`

Alejandro-Vega (changes requested):
Super small change requested. Everything else LGTM! Well done.

src/components/DataSubmissions/DataSubmissionListFilters.tsx
amattu2: Thanks for pointing this out. While testing this change, I also noticed that an empty selection is not persisted through page reloads. Even though it's arguably useless to support this use case, I decided to just fix it.

Logic simplified and above issue fixed in 368abe0. LMK if you have any concerns with these changes.

Alejandro-Vega (approved):
LGTM!

PR #608, opened 2025-01-27
CRDCDH-2190 Model Navigator – TSV, JSON, and PDF Dictionaries

amattu2:
### Overview

This PR updates our version of Data Model Navigator to support new Data Dictionary exports:

- TSV
- JSON

Additionally, for both the above types, and the existing PDF type, we now support exporting of only Required properties, or all of them.

> [!NOTE]
> The styling between the Figma mockup and the actual implementation differs slightly due to integration issues between DMN and CRDC. The design may not be 100% finalized yet, so there may be another PR in the future if needed.

- See Model Navigator changes here: https://github.com/CBIIT/Data-Model-Navigator/compare/84e0ab1c4238e1b6c89314f4af959a7af0fba511...b5d4f1873e646784b35dcc7a3c8eebc83f768bb3

### Related Ticket(s)

CRDCDH-2190 (Primary task)
CRDCDH-2191 (Filename task)
CRDCDH-2069 (Userstory)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8897,7 +8897,7 @@
     },
     "node_modules/data-model-navigator": {
       "version": "1.1.34",
-      "resolved": "git+ssh://git@github.com/CBIIT/Data-Model-Navigator.git#5a3aa718b511b70320e598bfa572731470e209cf",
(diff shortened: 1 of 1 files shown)


Alejandro-Vega (approved):
Looks really great and feels smooth. Only issue I could find was that the file name is slightly different from the US. You include "_v5.0.4", but the US just uses "_5.0.4". Although, I think yours is better and more descriptive.

Will approve for now, this can be addressed with your future PR.

PR #617, opened 2025-02-05
CRDCDH-2267 Submission Request version tracking

Alejandro-Vega:
### Overview

Updated Submission Request List table to include versioning.

> [!WARNING]
> Currently sorting by the version column is broken. The BE claims that it will be resolved when the real SR versions are restored. ([CRDCDH-2309](https://tracker.nci.nih.gov/browse/CRDCDH-2309))

- Retrieved the version from the application, and extracted the major and minor versions to ensure it is in the correct format
- Displayed the version in the Submission Request List table as a column

### Related Ticket(s)

[CRDCDH-2267](https://tracker.nci.nih.gov/browse/CRDCDH-2267) (Task)
[CRDCDH-2205](https://tracker.nci.nih.gov/browse/CRDCDH-2205) (US)

### DIFF
--- a/src/content/questionnaire/ListView.tsx
+++ b/src/content/questionnaire/ListView.tsx
@@ -7,7 +7,7 @@ import { useLazyQuery, useMutation } from "@apollo/client";
 import bannerSvg from "../../assets/banner/submission_banner.png";
 import { ReactComponent as BellIcon } from "../../assets/icons/filled_bell_icon.svg";
 import PageBanner from "../../components/PageBanner";
-import { FormatDate, Logger } from "../../utils";
+import { extractVersion, FormatDate, Logger } from "../../utils";
 import { useAuthContext, Status as AuthStatus } from "../../components/Contexts/AuthContext";
 import {
   LIST_APPLICATIONS,
@@ -167,6 +167,11 @@ const columns: Column<T>[] = [
     sx: {
       width: "124px",
     },
+  },
+  {
+    label: "Version",
+    renderValue: (a) => extractVersion(a.version) || "",
+    field: "version",
   },
   {
     label: "Submitted Date",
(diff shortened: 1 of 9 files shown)


src/content/questionnaire/ListView.tsx
```diff
@@ -168,6 +168,11 @@ const columns: Column<T>[] = [
       width: "124px",
     },
   },
+  {
```
amattu2, line 171: Just mentioning this since it's a good opportunity to bring it up – We seriously need to scope out how we can improve the rerendering of the GenericTable. While I was testing the error handling, I found that the `renderValue` function seems to be called x4 times on load. 

Most of our `renderValue` functions have low computation cost, but it's still extremely inefficient.

Screenshot 2025-02-06 at 9 31 01 AM

amattu2 (approved):
Everything FE looks good to me! I just annotated something about the GenericTable, but it is way out of the scope of this task, and something we need to address in a different release.

---

BTW, I'm unsure if I was doing something wrong, or if it just isn't working at all – But the BE never seems to upgrade the versions. I manually set one to 1.0 and saved it several times, it never upgraded to 3.0.

Merging as that's a BE issue.

src/content/questionnaire/ListView.tsx
amattu2: Resolving this so I can merge. We need to spend some time thinking about how we can improve the GenericTable so we have a game plan for when we have the allocation to actually refactor it.

amattu2: Actually it just occurred to me, in the prod build it's probably only x2 rerenders. StrictMode intentionally calls hooks twice. Regardless, we should try to get it down to x1.

PR #618, opened 2025-02-06
CRDCDH-2276 PBAC - Data Submission Cancel permission

Alejandro-Vega:
### Overview

Added the Data Submission Cancel permission and separated it from the Create permission.

> [!NOTE]
> The BE is not yet done so the new permission will not show in the Edit User page. Also, currently, the Cancel button still relies on the Create permission to Cancel from the BE.

- Defined new Cancel permission
- Updated Cancel button to use the new Cancel permission instead of the Create permission

### Related Ticket(s)

[CRDCDH-2276](https://tracker.nci.nih.gov/browse/CRDCDH-2276) (Task)
[CRDCDH-2215](https://tracker.nci.nih.gov/browse/CRDCDH-2215) (US)


src/config/AuthPermissions.ts
```diff
@@ -164,6 +164,7 @@ export const PERMISSION_MAP = {
 
       return false;
     },
+    cancel: NO_CONDITIONS,
```
amattu2, line 167: I can't test this scenario yet (discussed in slack), but wouldn't NO_CONDITIONS mean that I could cancel any Data Submissions that I can see? i.e. If I'm NOT a collaborator for your submission, I can still cancel it.

Technically this is the correct interpretation of requirement 3, but I don't think it's written correctly? I could be 100% wrong here.

amattu2 (changes requested):
One question annotated, otherwise it LGTM.

amattu2 (approved):
LGTM!

PR #628, opened 2025-02-14
fix: ErrorDetailsDialog should scroll on page body, not dialog body

amattu2:
### Overview

Fixes an issue where, when the content of the Error details dialog is very tall, the dialog shows a scrollbar inside itself instead of using the page body to scroll. 

This doesn't need to be merged before the demo.

N/A

### Related Ticket(s)

CRDCDH-2149


Alejandro-Vega (approved):
LGTM!

PR #641, opened 2025-03-06
CRDCDH-2418 Update Submission Requests URL

amattu2:
### Overview

This PR renames the Submission Request URL from "/submission[s]" to "/submission-request[s]", and provides a SEO-friendly redirect within our Nginx config.

> [!Warning]
> Some notes about the Nginx redirect:
>
> - The redirect only works when navigating from an external browser action, such as a bookmark or link from another application. Any **outdated** links from within our application will result in a 404 page, but I think I updated every outdated reference.
> - The redirect will forward any query parameters for the list page only (e.g. applied filters). I didn't forward the query params on the individual submission-request URL since I don't think we use them there.

- Rename `/submissions` and `/submission/{_ID}` to `/submission-requests` and `/submission-request/{_ID}`, respectively
- Create SEO-friendly (HTTP 301) redirects from the old URLs to the new ones
- Update all application-level references to the outdated URLs
- Remove the homepage redirect to submission requests after login (I don't think it was ever supposed to do that?)

### Related Ticket(s)

CRDCDH-2418 (FE Task)
CRDCDH-2385 (US)

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -3,6 +3,16 @@ server {
      server_name  localhost;
      root   /usr/share/nginx/html;
 
+    # 3.3.0 MIGRATION: Redirect /submissions to /submission-requests
+    location /submissions {
+        return 301 /submission-requests$is_args$args;
+    }
+    
+    # 3.3.0 MIGRATION: Redirect /submission/:uuid to submission-request/:uuid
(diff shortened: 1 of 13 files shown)


Alejandro-Vega (approved):
LGTM! I wasn't able to get the nginx redirect to work on my side, but it looks correct and didn't break anything. Everything else looks solid!

PR #658, opened 2025-03-31
CRDCDH-2471 Support Export of Original Metadata Files

amattu2:
### Overview

This PR adds support for the downloading of original metadata files uploaded to a Data Submission.

> [!NOTE]
> The downloaded files currently do not conform to the naming convention in the user story. If BE doesn't want to address this, then there will be another PR to handle this in the FE. 

- Update the Data Submission Batch File List dialog to allow downloading of individual metadata files or all files
- Fix a long-standing issue with the pluralization of the file count in the dialog (e.g. "1 FILES")
- Fix an issue with the File List table footer showing 26px of the background color
- Migrate the File List dialog base components to our pre-styled dialog components
- Update test/storybook coverage for these changes
- Update the base storybook config to implement the Notistack provider (no alerts were showing up in stories)

### Related Ticket(s)

CRDCDH-2471 (Task)
CRDCDH-1211 (US)


src/components/GenericTable/index.stories.tsx
```diff
@@ -14,6 +14,11 @@ const meta: Meta<typeof GenericTable> = {
         <Story />
       </SearchParamsProvider>
     ),
+    (Story) => (
+      <div style={{ background: "#f2f2f2" }}>
```
Alejandro-Vega, line 18: I see that the tables now have some background color showing on the edges of the pagination, was this always there and just relied on the background color of the wrapper?

Alejandro-Vega (approved):
Everything looks solid. Your storybook of the Batch File List component looks great! Just had one question below, but no changes necessary. Will merge as-is.

src/components/GenericTable/index.stories.tsx
amattu2: @Alejandro-Vega Yeah, this issue was always present but we never noticed it because the wrapper always had a background specified (or the page background is already white). I added the story decorator to make this issue obvious for the future, but I was hesitant to address the root issue since it might have some other side-effect.

PR #662, opened 2025-04-04
CRDCDH-2477 Request Access with Institution selection

amattu2:
### Overview

Includes changes to the Request Access implementation to add an institution field.

> [!NOTE]
> This dialog still has severe `act` issues whenever I write coverage for it, so most of the tests are still commented out. Maybe Vitest will be less problematic with this issue. 

- Add a freesolo, autocomplete Institution field to the Request Access component
- Update stories and tests for the related changes

### Related Ticket(s)

CRDCDH-2477 (Task)
CRDCDH-2448 (US)


src/components/AccessRequest/FormDialog.tsx
```diff
@@ -109,22 +110,31 @@ const FormDialog: FC<Props> = ({ onClose, ...rest }) => {
     }
   );
 
+  const { data: listInstitutions } = useQuery<ListInstitutionsResp, ListInstitutionsInput>(
```
Alejandro-Vega, line 113: I believe the `InstitutionListContext` does the same API request as it does here. Could we use that here instead? Otherwise, if you are using it differently then feel free to disregard and keep this as-is.

Alejandro-Vega (changes requested):
One small suggestion below. The only other thing is what you already mentioned to the BE team with the `institutionName_String_maxLength_100` type error, so I was not able to submit. Everything else LGTM!

Alejandro-Vega (approved):
LGTM!

PR #668, opened 2025-04-08
CRDCDH-2610 Update display order

amattu2:
### Overview

Resort the Data Commons list based on new requirements:

- GC always first
- Test MDF always last
- Alphabetically otherwise

- Resort the Data commons and update the test to match new requirements
- Fix the test coverage for the DC context (When I changed the DataCommonContext props, I forgot to update the test but it went unnoticed because the `displayName` and `name` of the datacommon at index 0 was identical)

### Related Ticket(s)

CRDCDH-2610 (Task)
CRDCDH-2609 (US)


Alejandro-Vega (approved):
LGTM! Didn't test locally.

PR #685, opened 2025-04-15
CRDCDH-2644 Cancel/Restore Reason Field Max Length

amattu2:
### Overview

PR to enforce maximum length on Reason field for the Cancel/Restore dialog. 

> [!NOTE]
> Mentioning this here for future reference – React Hook Form uses async validation behind the scenes, which **I think** is the root cause for a lot of our "act" issues. Their suggestion is to use one "findBy..." query in the tests to ensure the async validation is complete before the test completes.
>
> https://react-hook-form.com/advanced-usage#TestingForm

- Enforce Reason field maxLength (HTML5 restriction)
- Constrain Reason field from 1-500 characters (RHF validation)
- Update tests for these changes

### Related Ticket(s)

CRDCDH-2644


Alejandro-Vega (approved):
LGTM, seems to be working. Also, thanks for the tip regarding the act issues. Will try this in the future.

PR #687, opened 2025-04-16
CRDCDH-2651 Bug: Fix reset button in column visibility popper

Alejandro-Vega:
### Overview

Previously, the "RESET" button in the column visibility popper just set all columns to visible. This was updated to instead reset to default.

- Fixed "RESET" button in the column visibility popper to reset to default based on columns that are hidden by default
- Fixed issue where deleting local storage value for column visibility caused site to crash

### Related Ticket(s)

[CRDCDH-2651](https://tracker.nci.nih.gov/browse/CRDCDH-2651)


amattu2 (approved):
LGTM

PR #699, opened 2025-05-01
CRDCDH-2481 Bug: Update study Primary Contact checkbox logic

Alejandro-Vega:
### Overview

The "Same as the Program Primary Contact" checkbox while creating/modifying a study did not allow for both the checkbox to be unselected and a Primary Contact to not be specified. This caused an issue where what the user saved and what was actually saved are different.

- Updated the "Same as the Program Primary Contact" checkbox to rely on the new useProgramPC property, but still taking into account the study being assigned to multiple programs being invalid scenario
- Updated queries, mutations, types, and tests regarding Approved Studies to better reflect what is actually being retrieved and used

### Related Ticket(s)

[CRDCDH-2481](https://tracker.nci.nih.gov/browse/CRDCDH-2481) (Bug)


amattu2 (approved):
LGTM!

PR #728, opened 2025-05-27
CRDCDH-2833 Maintenance Mode Page

amattu2:
### Overview

Two approaches:

- Redirect to maintenance page – Current issue: Login won't work
- Override content with maintenance page – Current issue: URL will change if user clicks on links

N/A

### Related Ticket(s)

N/A


src/router.tsx
```diff
@@ -30,130 +32,139 @@ const routes: RouteObject[] = [
...
+                component={<OperationDashboard />}
+                redirectPath="/operation-dashboard"
+                redirectName="Operation Dashboard"
+              />
+            ),
+          },
+        ],
+      },
+      {
+        path: "/maintenance",
```
Alejandro-Vega, line 166: Currently, this path is always visible, even when maintenance mode is not enabled. When the user is redirected to this page, there is no way for the user to know when maintenance is over by clicking anything on the screen or refreshing. The only way to check is by manually updating the URL to remove the "/maintenance" and checking if this works.

My suggestion would be to instead redirect to home page or show 404 page if maintenance mode is off. Unfortunately, to do so you may have to convert this object into a function and pass the isMaintenanceMode boolean as a param, then dynamically include the route. Which also means you'd have to make the maintenance call even sooner, within the App.tsx to access.

Since maintenance mode is probably not used often, this may not be worth the effort. LMK your thoughts, and disregard if you deem it not necessary.

Alejandro-Vega (changes requested):
The maintenance page looks great! I included just one concern below. Everything else looks solid.

src/router.tsx
amattu2: I do agree with this concern, specifically, the fact that users are always stuck on the maintenance page. I didn't implement the change exactly as suggested, but instead:

> If maintenance mode is on, **replace** the page content with the Maintenance Page instead of **redirecting**. If it's off, just render the outlet as normal. 

This would allow them to reload the page to recheck the status. I tried a few different approaches during initial development, and each of them had their own nuanced issues. Let me know if this approach is any better.

Alejandro-Vega: Your suggestion works better. Nice job!

Alejandro-Vega (approved):
LGTM!

PR #737, opened 2025-06-04
CRDCDH-2867 Bug: Fix select dropdown menu sizes and tooltip sizes

Alejandro-Vega:
### Overview

Provided a way to specify when the text inside of tooltips is dynamic (text is not provided by us). When this is enabled, the max width of the tooltip is increased by ~2x. Also, looked around the site and fixed any select/autocomplete that is causing the dropdown menu to have mismatched width with their inputs.

- Added a prop to StyledTooltip, dynamic, to be used when the text inside of tooltips is not static. When it is dynamic, the max width of the tooltip is increased to support the potential size of the text. 
  - The reason I did this instead of just increasing max width for all, is because all of the static hard-coded tooltips that looked weird after the increase, and would have all of the text in a single super long line.
- Fixed select/autocomplete inputs which have a dropdown menu that is too large or too small. Updated it to dynamically detect the right size, only when the dropdown opens.

> [!NOTE]
> The select/autocomplete dropdown menu solution is kinda hacky, but seems to work well. Open to alternative suggestions. Maybe in the future we can just apply a fix in one location instead of having to add it in each usage, but I didn't want to modify the styled component since we try to leave those pretty bare-bones.

### Related Ticket(s)

[CRDCDH-2867](https://tracker.nci.nih.gov/browse/CRDCDH-2867) (Bug)

### DIFF
--- a/src/components/AdminPortal/Studies/ApprovedStudyFilters.tsx
+++ b/src/components/AdminPortal/Studies/ApprovedStudyFilters.tsx
@@ -3,7 +3,7 @@ import { debounce, sortBy } from "lodash";
 import { Box, FormControl, MenuItem, Stack, styled } from "@mui/material";
 import { Controller, useForm } from "react-hook-form";
 import StyledOutlinedInput from "../../StyledFormComponents/StyledOutlinedInput";
-import BaseSelect from "../../StyledFormComponents/StyledSelect";
+import StyledSelect from "../../StyledFormComponents/StyledSelect";
 import { useSearchParamsContext } from "../../Contexts/SearchParamsContext";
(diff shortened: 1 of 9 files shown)


amattu2 (approved):
LGTM and works better than before. There's still instances where we simply cannot display the full name in the tooltip without overflowing, but I have no idea how we could solve this quickly. e.g.

Screenshot 2025-06-05 at 9 58 47 AM

---

Thanks for fixing the selectbox width issue, your solution looks better than just letting it consume the full screen width. I do agree with you/copilot that we absolutely need to abstract this logic somewhere (probably within the base component), but we don't have time to address that now.

Merging as-is. Nice job

PR #742, opened 2025-06-16
fix: Vitest collecting coverage against stories

amattu2:
### Overview

This PR fixes the fact that Vitest (and probably Jest was too) computing coverage against our stories. Our coverage jumped about 5% by ignoring those files – which cannot have test coverage written for it anyway. Additionally, we're now ignoring these files completely:

- Test utils
- Vitest setup
- Production conf files (not related to app code)
- Static public folder – See note below

Technically the `public` folder contains JS files that should be covered, but in the future we probably should move those out of the `public` folder and import them in our app directly. 

Related to #734

N/A

### Related Ticket(s)

N/A

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -4,6 +4,15 @@ import react from "@vitejs/plugin-react";
 import viteTsConfigPaths from "vite-tsconfig-paths";
 import svgr from "vite-plugin-svgr";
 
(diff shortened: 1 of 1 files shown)


Alejandro-Vega (approved):
LGTM!

PR #751, opened 2025-06-27
CRDCDH-2898 Data Explorer - TSV Export

amattu2:
### Overview

This PR builds on the Data Explorer - Study View by enabling users to download a TSV of the metadata for the selected node type. 

Notably, it introduces a new utility `fetchAllData` that ingests a LazyQueryExecFunction (i.e. the response from `useLazyQuery`) and fetches all of the data from an API using a paginated approach, rather than querying for all of the data at once. Very large datasets cause significant issues in the BE when querying for everything at once, so we'll need to use this approach going forward with other export buttons.

> [!NOTE]
> For testing large batch exports, use `ALECS-DEMO-STUDY` and `study_site` node for ICDC.

- Add a export button to the study view table
- Update the Data Explorer filters component to allow customization of visible actions (rather than prop drill a ton of info)
- Add a new utility function to aggregate query responses into one dataset
- Update `listReleasedDataRecords` query to include the `properties` param

### Related Ticket(s)

CRDCDH-2898 (Task)
CRDCDH-2828 (US)


Alejandro-Vega (approved):
Seems to work well! As you mentioned in your TODO, it could probably benefit greatly from parallel calls. Does not need to be addressed here.

PR #779, opened 2025-07-15
CRDCDH-2902

amattu2:
### Overview

N/A

N/A

### Related Ticket(s)

CRDCDH-2902 (Task)

### DIFF
--- a/src/components/ErrorDetailsDialog/index.stories.tsx
+++ b/src/components/ErrorDetailsDialog/index.stories.tsx
@@ -1,11 +1,4 @@
-import { MockedResponse } from "@apollo/client/testing";
 import type { Meta, StoryObj } from "@storybook/react";
-
-import {
-  RETRIEVE_RELEASED_DATA,
-  RetrieveReleasedDataInput,
-  RetrieveReleasedDataResp,
-} from "../../graphql";
 
 import Dialog from "./index";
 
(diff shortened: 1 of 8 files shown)


src/components/ErrorDetailsDialog/v2/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+    </StyledHeader>
+    <StyledPostHeader variant="caption" data-testid="error-details-node-info">
+      {postHeader}
+    </StyledPostHeader>
+    <StyledIssueCount variant="body2" data-testid="error-details-error-count">
+      {issues?.length || 0} {issues?.length === 1 ? "ISSUE" : "ISSUES"}:
+    </StyledIssueCount>
+    <StyledList>
+      {issues.map((issue, idx) => (
+        <StyledIssueContainer key={issue.message} data-testid={`error-details-issue-${idx}`}>
```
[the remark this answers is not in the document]
amattu2: Using index isn't a stable alternative. I went with severity+message to reduce the odds of a conflict. I'm not sure if duplicate messages can even appear anyway. d507ec9

src/components/ErrorDetailsDialog/v2/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+  color: "#595959",
+  fontFamily: "'Nunito', 'Rubik', sans-serif",
+  fontSize: "16px",
+  fontStyle: "normal",
+  fontWeight: 400,
+  lineHeight: "19.6px",
+});
+
+const StyledList = styled(List)({
+  overflowY: "auto",
```
Alejandro-Vega, line 56: Is now a good time to switch over to scroll body instead? Disregard, if the scrolling is specified in design

src/components/PVRequestButton/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+  const [requestPV] = useMutation<RequestPVResponse, RequestPVInput>(REQUEST_PV, {
+    context: { clientName: "backend" },
+    fetchPolicy: "no-cache",
+  });
+
+  const [loading, setLoading] = useState<boolean>(false);
+  const [confirmOpen, setConfirmOpen] = useState<boolean>(false);
+
+  const confirmButtonProps = useMemo<ButtonProps>(
```
Alejandro-Vega, line 105: I see that in the Figma design the "Confirm and Submit" button is the info color, but here we have 2 separate buttons. I'd suggest verifying the desired color for this button, I think it would make more sense to use variant contained and color "success" like in the Access Request Dialog. Seems inconsistent to use this info color here.

src/components/PVRequestButton/index.tsx
```diff
@@ -0,0 +1,252 @@
...
+  };
+
+  const onCloseDialog = async () => {
+    setConfirmOpen(false);
+  };
+
+  const onConfirmDialog = useCallback(async () => {
+    setLoading(true);
+    try {
+      const comment = watch("comment")?.trim();
```
Alejandro-Vega, line 125: I think it would be cleaner to use handleSubmit(onConfirmDialog) and get the form values through the params. LMK if there is a performance issue with this, and instead your way is better.

Alejandro-Vega (changes requested):
Looks great and works well! Small requested changes below.

amattu2:
Updated.

FYI The API doesn't seem to work ATM because of a feature branch being deployed.

Alejandro-Vega (approved):
LGTM! Didn't retest. Due to the API failing, feel free to merge whenever you feel is right.

PR #797, opened 2025-07-21
CRDCDH-2990 Manage Collaborators with 'No Access'

Alejandro-Vega:
### Overview

Update the manage collaborators table to re-add the "Access" column. 

- Re-added the "Access' column within the manage collaborators table
- Kept the option "Can Edit", and added "No Access"
- Added test and storybook coverage

### Related Ticket(s)

[CRDCDH-2990](https://tracker.nci.nih.gov/browse/CRDCDH-2990) (Task)
[CRDCDH-2693](https://tracker.nci.nih.gov/browse/CRDCDH-2693) (US)

### DIFF
--- a/src/components/Collaborators/CollaboratorsTable.tsx
+++ b/src/components/Collaborators/CollaboratorsTable.tsx
@@ -31,6 +31,11 @@ const StyledTableContainer = styled(TableContainer)(() => ({
   marginBottom: "15px",
 }));
 
+const FixedTable = styled(Table)({
+  tableLayout: "fixed",
+  width: "100%",
+});
+
 const StyledTableHeaderRow = styled(TableRow)(() => ({
   "&.MuiTableRow-root": {
     height: "38px",
@@ -197,7 +202,13 @@ const CollaboratorsTable = ({ isEdit }: Props) => {
   return (
     <>
       <StyledTableContainer data-testid="collaborators-table-container">
-        <Table>
+        <FixedTable>
+          <colgroup>
+            <col style={{ width: isEdit ? "47%" : "50%" }} />
+            <col style={{ width: isEdit ? "40%" : "50%" }} />
+            {isEdit && <col style={{ width: "13%" }} />}
+          </colgroup>
+
           <TableHead>
             <StyledTableHeaderRow data-testid="table-header-row">
               <StyledTableHeaderCell id="header-collaborator" data-testid="header-collaborator">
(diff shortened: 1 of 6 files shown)


src/components/Collaborators/CollaboratorsTable.tsx
```diff
@@ -197,7 +260,7 @@ const CollaboratorsTable = ({ isEdit }: Props) => {
                     renderValue={() => (
                       <TruncatedText
                         text={collaborator.collaboratorName ?? " "}
-                        maxCharacters={35}
+                        maxCharacters={20}
                         underline={false}
```
amattu2, line 264: Do we need to drop the truncation this low? It seems like we have a little more room.

amattu2 (changes requested):
Left one comment. Aside from that, it seems to work well.

src/components/Collaborators/CollaboratorsTable.tsx
```diff
@@ -216,6 +279,48 @@ const CollaboratorsTable = ({ isEdit }: Props) => {
                       ))}
                   </StyledSelect>
                 </StyledNameCell>
+
+                <StyledTableCell data-testid={`collaborator-access-${idx}`}>
```
amattu2, line 283: One additional change:

The user story was just updated to clarify the implementation. The Access column is not a radio group (like before), it's just text with the access level displayed.

src/components/Collaborators/CollaboratorsTable.tsx
Alejandro-Vega: Unfortunately, it looks like a lot of space is available because I do my testing with wider characters like "A". So something like "John" is less wide than "AAAA". We should find a better solution in the future for this.

Screenshot 2025-07-30 at 2 39 39 PM
Screenshot 2025-07-30 at 2 40 49 PM

src/components/Collaborators/CollaboratorsTable.tsx
```diff
@@ -153,12 +158,22 @@ const CollaboratorsTable = ({ isEdit }: Props) => {
...
+            <col style={{ width: isEdit ? "36%" : "50%" }} />
+            {isEdit && <col style={{ width: "14%" }} />}
+          </colgroup>
+
           <TableHead>
             <StyledTableHeaderRow data-testid="table-header-row">
               <StyledTableHeaderCell id="header-collaborator" data-testid="header-collaborator">
                 Collaborator
               </StyledTableHeaderCell>
+              <StyledTableHeaderCell id="header-access" data-testid="header-access">
```
amattu2, line 173: Unless this width was specified somewhere or requested by KC, let's drop the size of the Access column and increase the truncation limit for the Collaborators column. Looks like the width is still leftover from when it was a radio.

Image

PR #803, opened 2025-07-25
Bug: Fix validation type default value causing Validation status to show In-Progress

Alejandro-Vega:
### Overview

Logic previously expected `validationType` to be null as a default value, but now it is an empty array (Due to BE Prisma change). This caused the validation status to show as In-Progress, when it shouldn't be shown. Updated to check array length instead.

N/A

### Related Ticket(s)

N/A


amattu2 (approved):
LGTM! Thanks for fixing this.

PR #805, opened 2025-07-30
CRDCDH-2995 Model Navigator – Use MDF CDE Data

amattu2:
### Overview

This PR refactors our integration with Model Navigator, specifically for the CDE portion.

Model Navigator Changes (https://github.com/CBIIT/Data-Model-Navigator/compare/ea0b4ffec31b0f6042e6d3f1c4e6568a6a9c01c0..84f96fa5fefae5c75b2b361a0748a1779f8c7034):

- Instead of depending on the integration (i.e. us) to provide CDE information, DMN now pulls this directly from the MDF
- Since a property can have multiple CDEs (`Term`), it handles the terms as an array and displays all terms
- Updated PDF, TSV, and JSON exports to match

Frontend Changes:

- Removed usage of `cdeMap` and just pulled the CDE information directly from the DMN state
- Since we don't want non-caDSR CDEs, we delete all non-caDSR options before passing the info back to DMN
- Rewrote most utilities for CDE handling, but kept the same logic for PVs
- Updated test coverage where needed
- Added factories for the MDFDictionary type
- Fixed the DMN storybook auto upgrading to the latest GC model version, which triggers Chromatic changes

### Related Ticket(s)

CRDCDH-2995
CRDCDH-2978


src/utils/dataModelUtils.ts
```diff
@@ -136,114 +134,118 @@ export const buildFilterOptionsList = (config: ModelNavigatorConfig): string[] =
...
+ *
+ * Key details on handling:
+ * 1. Updates the Value field (CDE Name)
+ * 2. Updates the property enum
+ * 3. Updates the property type
+ *
+ * @param dictionary the MDF dictionary to populate with data
+ * @param data the API response data
+ */
+export const populateCDEData = (
```
Alejandro-Vega, line 175: Similarly as the deleteInvalidCDEs function, we can refactor this with a flatMap. Also, if the CDECode + CDEVersion is unique, we could do a mapping of the CDECode:CDEVersion to make a quick look-up instead of repeatedly traversing the entire data looking for the matching code and version. Feel free to adjust/change anything, especially if I missed something.

```typescript
export const populateCDEData = (
  dictionary: MDFDictionary,
  data: RetrieveCDEsResp["retrieveCDEs"]
): void => {
  if (!dictionary || Object.keys(dictionary).length === 0) {
    return;
  }

  const cdeMap = new Map<string, (typeof data)[number]>();
  data?.forEach((d) => cdeMap.set(`${d.CDECode}::${d.CDEVersion}`, d));

  const HELP_DESK_MSG =
    "Permissible values are currently not available. Please contact the CRDC Submission Portal HelpDesk at NCICRDCHelpDesk@mail.nih.gov";

  Object.values(dictionary)
    .flatMap((node) => Object.values(node?.properties ?? {}))
    .forEach((prop) => {
      if (!Array.isArray(prop?.Term)) {
        return;
      }

      prop.Term.forEach((term) => {
        if (!term?.Code || !term?.Version) {
          return;
        }

        const apiData = cdeMap.get(`${term.Code}::${term.Version}`);

        if (!apiData && prop.enum) {
          Logger.error("dataModelUtils: Unable to match CDE for property", prop, term);
          prop.enum = prop.enum ? [HELP_DESK_MSG] : undefined;
          return;
        }

        if (!apiData) {
          // Do nothing?
          return;
        }

        // Update CDE Name
        term.Value = apiData.CDEFullName;

        // Populate Permissible Values if available from API
        if (Array.isArray(apiData.PermissibleValues) && apiData.PermissibleValues.length > 0) {
          prop.enum = apiData.PermissibleValues;
        } else if (
          Array.isArray(apiData.PermissibleValues) &&
          apiData.PermissibleValues.length === 0 &&
          prop.enum
        ) {
          // Permissible Values from API are empty, convert property to "string" type
          delete prop.enum;
          prop.type = "string";
        }
      });
    });
};
```

src/utils/dataModelUtils.ts
```diff
@@ -136,114 +134,118 @@ export const buildFilterOptionsList = (config: ModelNavigatorConfig): string[] =
...
-        apiError,
-        `${parentKey}.${subKey}`
-      );
+/**
+ * A utility to extract all unique CDEs from the MDF dictionary.
+ *
+ * @param dictionary the MDF dictionary to extract CDEs from
+ * @return An array of objects containing the CDE Code/Version/Origin
+ */
+export const extractAllCDEs = (dictionary: MDFDictionary): CDEInfo[] => {
```
Alejandro-Vega, line 230: This can also be cleaned up with a flatMap. Feel free to follow the same structure as the others. Although, while looking through [Lodash docs](https://lodash.com/docs/#chain), I found they have an interesting chain wrapper. Unsure of its performance or how I feel about it, but it looks cool. Just thought I'd bring it up as an option, but feel free to disregard it if you don't want to use it. Here's an example:

```typescript
export const extractAllCDEs = (dictionary: MDFDictionary): CDEInfo[] => {
  if (!dictionary || Object.keys(dictionary).length === 0) {
    return [];
  }

  return chain(dictionary)
    .values()
    .flatMap((node) => values(node?.properties ?? {}))
    .flatMap("Term")
    .filter((t) => t?.Code && t?.Version && t?.Origin)
    .uniqBy((t) => `${t.Code}:${t.Version}:${t.Origin}`)
    .map((t) => ({
      CDECode: t.Code,
      CDEVersion: t.Version,
      CDEOrigin: t.Origin,
    }))
    .value();
};
```

Alejandro-Vega (changes requested):
Great work, it seems to work super well. I left some optional comments on how to refactor to avoid the nesting. Feel free to disregard any or all of them.

src/utils/dataModelUtils.ts
amattu2: Thanks for the suggestions. I simplified the CDE lookup and the iteration logic, but I left everything else related to the enum handling alone.

788945a

PR #818, opened 2025-08-11
CRDCDH-3149 [Bug]: Fix Release button being enabled when cross validation had errors

Alejandro-Vega:
### Overview

There was an edge-case where when `crossValidationStatus` was "Error", but there was a related "In-Progress" submission, then it would short circuit to enable the Release button with alert. Updated the order of the logic to fix this.

- Updated Release button logic so if `crossValidationStatus` is ever "Error", then it is disabled no matter what
- Rearranged logic of Scenario 2 and 3, since checking for Submitted/Released was being done twice
- Removed "Passed" short-circuit due to it not showing alert when there are related "In Progress" submissions. Instead I let it go through each scenario to ensure correct display of button and alert
- Added test cases for scenarios that caused the above issues

### Related Ticket(s)

[CRDCDH-3149](https://tracker.nci.nih.gov/browse/CRDCDH-3149) (Bug)

### DIFF
--- a/src/utils/dataSubmissionUtils.test.ts
+++ b/src/utils/dataSubmissionUtils.test.ts
@@ -977,26 +977,6 @@ describe("shouldDisableRelease", () => {
     }
   );
 
-  it.each<CrossSubmissionStatus>(["Passed"])(
-    "should allow release when crossSubmissionStatus is %s even if other submissions exist",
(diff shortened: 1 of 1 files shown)


amattu2 (approved):
LGTM, but I didn't test the scenario locally.

PR #834, opened 2025-08-21
Relative CI - Production Bundle Analysis

amattu2:
### Overview

N/A

N/A

### Related Ticket(s)

N/A


amattu2:
Need to rename the action because it's showing up as "Build" and that's confusing with "Build Frontend Service"

PR #844, opened 2025-08-25
CRDCDH-3107 [Bug]: No unsaved modal after canceling SR with unsaved changes

cywu567:
### Overview

If a user edits the SR and cancels it, users undergo the canceling procedures and then are taken directly to the SR list without seeing the unsaved changes dialog.

- Used a ref to track cancel state
- Updated navigation blocker and unload handler to bypass the unsaved changes dialog after canceling a SR

### Related Ticket(s)

[CRDCDH-3107](https://tracker.nci.nih.gov/browse/CRDCDH-3107)


Alejandro-Vega (approved):
LGTM! Seems to work well. Nice job!

PR #869, opened 2025-09-19
CRDCDH-3263 Add a FAQ for pasting into fields

amattu2:
### Overview

PR to address pasting into excel by adding a note on how to do it. I'm open to other verbiage on the note if it isn't clear. 

> [!NOTE]
> This potentially could be solved by adding worksheet protections, but it's super tedious. I left more details in the bug.

N/A

### Related Ticket(s)

CRDCDH-3263

### DIFF
--- a/src/classes/Excel/Instructions/Content.ts
+++ b/src/classes/Excel/Instructions/Content.ts
@@ -59,6 +59,12 @@ export const CONTENT = {
           answer: "Dates should be entered in the format MM/DD/YYYY.",
           rowHeight: 20,
         },
+        {
+          question: "Q: Can I paste values directly into the spreadsheet fields?",
+          answer:
+            "Yes, you can paste values directly into the spreadsheet fields, but please note that pasting content may remove dropdown menus and special features in some cells. For best results, use 'Paste Special' and select 'Values', or enter your data manually.",
+          rowHeight: 60,
+        },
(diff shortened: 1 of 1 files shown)


src/classes/Excel/Instructions/Content.ts
```diff
@@ -59,6 +59,12 @@ export const CONTENT = {
           answer: "Dates should be entered in the format MM/DD/YYYY.",
           rowHeight: 20,
         },
+        {
+          question: "Q: Can I paste values directly into the spreadsheet fields?",
+          answer:
```
Alejandro-Vega, line 64: I think starting the answer with "Yes" might lead users to read that and ignore the rest. I'd suggest starting it with something like "Pasting into the Excel form may lead to unexpected results such as ... " then later add something like "To safely paste into the form, ...". You don't have to use my words, I am just telling you where my mind is at with this.

Alejandro-Vega (changes requested):
One small optional suggestion below. Otherwise, LGTM.

src/classes/Excel/Instructions/Content.ts
Alejandro-Vega: All 3 options are solid. I think 1 is the most straight-forward. I'll leave the choice to you.

PR #871, opened 2025-09-25
CRDCDH-3273 Fix dynamic snackbar text flowing outside of alert region

amattu2:
### Overview

PR to fix situations where the content rendered in the snackbar does not break a very long word, and so the text flows outside of the alert element.

Also created a storybook for notistack.

N/A

### Related Ticket(s)

CRDCDH-3273 (Bug)


Alejandro-Vega (approved):
LGTM!

PR #903, opened 2025-11-24
CRDCDH-3323 [Bug]: 508 issues

Alejandro-Vega:
### Overview

Fixed broken aria reference and missing header accessibility issues.

- Fixed broken aria reference with SR "Start a Submission Request" acknowledgement dialog
- Fixed missing header accessibility error by providing aria label. This was due to Dialog does not have a header

False-positives that could not be resolved:
- All MUI select dropdowns are now throwing contrast error on native element. This is a false-positive, and couldn't find an easy fix
- All MUI Textarea elements throw "Missing form label" accessibility errors. This is also a false-positive, MUI has 2 input elements in their textarea component.

### Related Ticket(s)

[CRDCDH-3323](https://tracker.nci.nih.gov/browse/CRDCDH-3323)

### DIFF
--- a/src/components/CreateApplicationButton/index.tsx
+++ b/src/components/CreateApplicationButton/index.tsx
@@ -140,6 +140,7 @@ const CreateApplicationButton: FC<CreateApplicationButtonProps> = ({
             },
           },
         }}
+        aria-labelledby="dialog-description"
       />
     </>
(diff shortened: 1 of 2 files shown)


amattu2 (approved):
LGTM

PR #913, opened 2025-12-04
fix: Unit testing in CI using legacy options from Jest

amattu2:
### Overview

N/A

N/A

### Related Ticket(s)

N/A


amattu2:
This saved almost no time and introduced timing-based issues. Not worth pursuing.

PR #923, opened 2025-12-10
fix: PDFs should not be included in the bundle

amattu2:
### Overview

N/A

N/A

### Related Ticket(s)

N/A


amattu2:
This is incorrectly inflating our bundle size. We will need to address this in the future.

PR #928, opened 2026-01-09
Bump Vulnerable Packages

amattu2:
### Overview

This isn't tied to a specific task but it resolves most (~5 remaining) of our vulnerable packages. DMN is blocking us from upgrading jsPDF and Axios.

I did a quick run through of the app and it seems to all work fine, but please also do a check to make sure nothing abnormal came up.

The only change I saw is that the router blocker is no longer considered unstable, so I updated the imports in our FormView.

- `npm audit fix`, non breaking upgrades only

### Related Ticket(s)

N/A


amattu2:
Well... This may not work after-all. 60 test failures is somewhat intense.

amattu2:
@copilot Debug all of the failing unit tests caused by dependency version bumping and fix them. Do NOT mock the console or try to work around the legitimate warning, we need to update the unit tests or router configuration to solve the root of the issue. Do not modify any sourcecode.

Alejandro-Vega:
> Well... This may not work after-all. 60 test failures is somewhat intense.

@amattu2 The failed tests seem to be warnings from react-router-dom. From the very few I tested, they can be mitigated by adding:
`future={{ v7_startTransition: true, v7_relativeSplatPath: true }}` prop to the MemoryRouter. I did not look into the downsides of using this, but it seems to make the tests pass. It might be worth looking into.

Alejandro-Vega:
Ah disregard above, I see you have a separate PR to handle those changes.

Alejandro-Vega (dismissed):
I have some slight hesitation, mainly for the package-lock changes such as bumping Vite, router, quicksight, etc. due to not knowing what those new versions do. I don't notice anything broken on the site, but will require further testing.

Letting you have the final say to merge this or not.

amattu2:
Given that we're both hesitant, I'm going to rebase this into 3.6.0 instead. We'll have a longer period to test the changes and ensure nothing odd is introduced.

PR #931, opened 2026-01-12
CRDCDH-3478 Increase DOI character limit to 200

amattu2:
### Overview

A PR to bump the DOI maximum length to 200 characters, from 20.

- Bump DOI length limit on Zod Schema and Section B Form
- Change the form placeholder to indicate the character limit
- Update test case for Excel parser only, we don't have unit tests for the SR Form

### Related Ticket(s)

CRDCDH-3478
CRDCDH-3477 (US)


Alejandro-Vega (approved):
LGTM! Didn't test locally.

PR #948, opened 2026-02-17
CRDCDH-3498 [Bug]: Questionnaire Excel Form dbGaP ID is invalid

Alejandro-Vega:
### Overview

There was an issue with the dbGaP ID field throwing an error during Zod validation within the Questionnaire import.

- Updated validation for dbGaPPPHSNumber to allow for empty string
- Added the super refine when validating section status. Without this, if you upload a valid form to complete the section, then upload an unfinished form, then it will remain as complete
- Stored validationIssues into the middleware to test for this specific case

**Note:** Unsure if other sections need the super refine within ApplicationSections. I originally added it to all of them, but it caused other issues. The rest would require further investigation. 

### Related Ticket(s)

[CRDCDH-3498](https://tracker.nci.nih.gov/browse/CRDCDH-3498) (Bug)

### DIFF
--- a/src/classes/QuestionnaireExcelMiddleware.test.ts
+++ b/src/classes/QuestionnaireExcelMiddleware.test.ts
@@ -1,4 +1,5 @@
 import ExcelJS, { Worksheet } from "exceljs";
+import { isEqual } from "lodash";
 import { v4 } from "uuid";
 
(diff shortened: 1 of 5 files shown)


amattu2 (approved):
LGTM

PR #963, opened 2026-03-18
CRDCDH-3563 [Bug]: Fix value not properly updating after import

Alejandro-Vega:
### Overview

When you import a value, then import again with a different data type/other data type, then it will not properly update. Fixed it to update when importing new value.

N/A

### Related Ticket(s)

[CRDCDH-3563](https://tracker.nci.nih.gov/browse/CRDCDH-3563) (Task)


amattu2 (approved):
LGTM. Did not test locally

PR #974, opened 2026-03-30
CRDCDH-3609 Enable/Disable ChatBot using config

Alejandro-Vega:
### Overview

Added the ability to enable/disable the ChatBot using query or env setup.

- The query needs to return enabled and the env var `VITE_CHATBOT_API_BASE_URL` needs to be defined for the chatbot to be displayed
- Added gate to prevent rendering of chatbot when not enabled or env var missing

### Related Ticket(s)

[CRDCDH-3609](https://tracker.nci.nih.gov/browse/CRDCDH-3609) (US)
[CRDCDH-3611](https://tracker.nci.nih.gov/browse/CRDCDH-3611) (Task)


src/components/ChatBotGate/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+/**
+ * Provides a rendering gateway for toggling the ChatBot feature.
+ * If the ChatBot API base URL is not configured or the ChatBot is disabled, it will render nothing.
+ * Otherwise, it will render the children.
+ *
+ * @returns The ChatBot gate component.
+ */
+const ChatBotGate: FC<Props> = ({ children }) => {
```
amattu2: Would it make sense to lazy-load the component from here instead of in the layout? If the chatbot is disabled, we can then skip those chunks completely.

amattu2 (dismissed):
LGTM. If the chatbot is disabled, or the env isn't configured, the chunk isn't even fetched. Very nice design!

Can be merged once ready.

PR #991, opened 2026-05-05
CRDCDH-3729 Fix template import running double migrations

amattu2:
### Overview

This PR fixes the SRF Template Import functionality running migrations twice when importing into a newly created SRF (i.e. one that isn't saved in the backend yet). The first migration is initiated by the SRF import, and causes _no issues_ because we explicitly disable autofilling of PI info, but the second migration occurs because of the URL change (`new` -> uuid) and does not explicitly disable the PI autofilling (not good).

N/A

### Related Ticket(s)

CRDCDH-3729


Alejandro-Vega (approved):
LGTM!

PR #995, opened 2026-05-06
CRDCDH-3734 [Bug]: Update date validation for planned publication expected date

Alejandro-Vega:
### Overview

Within the SRF within the planned publications, the expected date was not properly validating future dates when importing a form. Updated to properly validate, but still persist value if validation fails

- Updated Zod schema for planned publications expected date to only allow same day or future dates
- Added planned publication expected date field to persist the value even if fails past date validation. Without this it would not update the SRF with the invalid date

### Related Ticket(s)

[CRDCDH-3734](https://tracker.nci.nih.gov/browse/CRDCDH-3734) (Bug)


amattu2 (approved):
LGTM!

